PR #209, opened 2022-12-01
[JIT-504] Add low balance check in uploading merkle roots

esemeniuc:
Notes:
- Fixes a bunch of clippy warnings

### DIFF
--- a/bench-batch-simulate-bundle/src/main.rs
+++ b/bench-batch-simulate-bundle/src/main.rs
@@ -232,7 +232,7 @@ fn spawn_slots_subscribe_thread(
     node_name: String,
     exit: Arc<AtomicBool>,
 ) -> JoinHandle<()> {
-    let mut slots_sub = PubsubClient::slot_subscribe(&*pubsub_addr).unwrap();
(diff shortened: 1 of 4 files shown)


bench-batch-simulate-bundle/src/main.rs
```diff
@@ -111,10 +111,14 @@ fn main() {
     let t_hdls = vec![
         spawn_slots_subscribe_thread(
             args.simulation_ws_url,
-            "simulation-node".into(),
+            "simulation-node".to_string(),
```
buffalu, line 114: drop these

tip-distributor/src/merkle_root_upload_workflow.rs
```diff
@@ -1,3 +1,5 @@
+use solana_program::fee_calculator::DEFAULT_TARGET_LAMPORTS_PER_SIGNATURE;
+use solana_program::native_token::LAMPORTS_PER_SOL;
```
buffalu, line 2: should be in the block below?

buffalu (approved):

PR #219, opened 2022-12-22
feat: collect balances only on locked accounts

### LINKED ISSUE #50: Bank::collect_balances collects balances for transactions that fail to lock
#### Problem
In Bank::collect_balances, we collect balances on accounts that failed to lock

#### Proposed Solution
We should only collect balances for transactions that have a corresponding Ok(()) on lock_results

anoushk1234:
Earlier balances were collected even on accounts that weren't locked. Here we check if the account is locked and only then update the balances. Fixes #50

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3317,6 +3317,15 @@ dependencies = [
 
 [[package]]
 name = "pbkdf2"
+version = "0.10.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


anoushk1234:
The tests fail because there's an assertion that checks if len of balances is 2, one of the accounts isn't locked so it returns 1, I'll change the test in the next commit to check eq to 1

anoushk1234:
Let me know if there should be any changes/improvements, specifically in the test case, thanks!

segfaultdoc:
PR moved [here](https://github.com/jito-foundation/jito-solana/pull/258)

PR #224, opened 2023-01-09
add new args to support single relayer and block-engine endpoints

segfaultdoc:
We now support a single DNS entry routing to the different block-engine services. Relayer will support the same Soon TM.

TODO:
- test new args

### DIFF
--- a/validator/src/main.rs
+++ b/validator/src/main.rs
@@ -1782,32 +1782,16 @@ pub fn main() {
                 .hidden(true),
         )
         .arg(
-            Arg::with_name("relayer_address")
-                .long("relayer-address")
-                .value_name("relayer_address")
+            Arg::with_name("block_engine_url")
+                .long("block-engine-url")
+                .help("Block engine url")
                 .takes_value(true)
-                .help("Address of the relayer")
         )
         .arg(
-            Arg::with_name("block_engine_address")
-                .long("block-engine-address")
-                .value_name("block_engine_address")
+            Arg::with_name("relayer_url")
+                .long("relayer-url")
+                .help("Relayer url")
                 .takes_value(true)
-                .help("Address of the block engine")
-        )
-        .arg(
-            Arg::with_name("block_engine_auth_service_address")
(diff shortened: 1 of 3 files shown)


validator/src/main.rs
```diff
@@ -1782,32 +1782,14 @@ pub fn main() {
...
-                .takes_value(true)
-                .help("Address of the relayer")
-        )
-        .arg(
-            Arg::with_name("block_engine_address")
-                .long("block-engine-address")
-                .value_name("block_engine_address")
-                .takes_value(true)
-                .help("Address of the block engine")
+            Arg::with_name("block_engine_url")
```
buffalu, line 1785: can you change to block_engine_address? then they can just remove one line from startup

validator/src/main.rs
```diff
@@ -1782,32 +1782,14 @@ pub fn main() {
...
-                .takes_value(true)
-                .help("Address of the block engine's authentication service.")
-        )
-        .arg(
-            Arg::with_name("relayer_auth_service_address")
-                .long("relayer-auth-service-address")
-                .value_name("relayer_auth_service_address")
-                .takes_value(true)
-                .help("Address of the block engine's authentication service.")
+            Arg::with_name("relayer_url")
```
buffalu, line 1790: can you change to relayer_address?

validator/src/main.rs
buffalu: and can deprecate the old ones

buffalu: nvmd. if remove one line but they keep the port, it'll try to auth through block engine specific port + fail

validator/src/main.rs
```diff
@@ -3513,6 +3526,10 @@ lazy_static! {
             "Vote account sanity checks are no longer performed by default.",
         ),
         ("no_rocksdb_compaction", ""),
+        ("block_engine_address", ""),
```
buffalu, line 3529: should add messages here to use the new arguments

PR #225, opened 2023-01-10
Buffer bundles that exceed cost model

jedleggett:
#### Problem
Bundles are failing the QOS cost model when blocks are full.  

#### Summary of Changes
Buffer bundles that exceed the QOS cost model.

### DIFF
--- a/core/src/bundle_stage.rs
+++ b/core/src/bundle_stage.rs
@@ -1016,6 +1016,8 @@ impl BundleStage {
                         bundle_stage_leader_stats
                             .bundle_stage_stats()
                             .increment_execution_results_exceeds_cost_model(1);
+                        // retry the bundle
+                        unprocessed_bundles.push_back(packet_bundle);
                     }
                     Err(BundleExecutionError::TipError(_)) => {
                         bundle_stage_leader_stats


core/src/bundle_stage.rs
```diff
@@ -322,6 +313,19 @@ impl BundleStage {
             ),
         );
```
jedleggett, line 314: early return needs to occur after metrics accumulation

core/src/bundle_stage.rs
```diff
@@ -304,16 +305,6 @@ impl BundleStage {
             &bank_start.working_bank,
         );
 
-        // qos rate-limited a tx in here, drop the bundle
-        if sanitized_bundle.transactions.len() != num_included {
```
buffalu, line 308: oh interesting, this looks like potential bug?

core/src/bundle_stage.rs
```diff
@@ -322,6 +313,19 @@ impl BundleStage {
             ),
         );
 
+        // either qos rate-limited a tx in here or bundle exceeds max cost, drop the bundle
+        let total_bundle_cost: u64 = tx_costs.iter().map(|tx_cost| tx_cost.sum()).sum();
+        if sanitized_bundle.transactions.len() != num_included
+            || total_bundle_cost > MAX_BUNDLE_COST
```
buffalu, line 319: what is MAX_BUNDLE_COST derived from?

core/src/bundle_stage.rs
buffalu: i wonder if we need to move the qos_service.report_metrics too?

core/src/bundle_stage.rs
buffalu: like where did that number come from?

core/src/bundle_stage.rs
```diff
@@ -1375,6 +1387,7 @@ impl BundleStage {
         let blacklisted_accounts = HashSet::from_iter([tip_manager.tip_payment_program_id()]);
 
         let mut unprocessed_bundles: VecDeque<PacketBundle> = VecDeque::with_capacity(1000);
+        let mut buffered_bundles: VecDeque<PacketBundle> = VecDeque::with_capacity(1000);
```
buffalu, line 1390: trying to think of better name for this...

core/src/bundle_stage.rs
jedleggett: right now i just chose something bigger than was possible with 5 txs,  this was mostly just to create a parameter that can be adjusted if we increase bundle size in the future.

should i comment or just remove and worry about that when we/if increase bundle size

core/src/bundle_stage.rs
jedleggett: wdym?

core/src/bundle_stage.rs
jedleggett: yes, good catch

core/src/bundle_stage.rs
jedleggett: cost_model_failed_bundles?

core/src/bundle_stage.rs
jedleggett: the early return before metrics report was definitely a bug.

core/src/bundle_stage.rs
buffalu: imo remove so we don't need to push out another update if we hard-code value or have validators set from cli

buffalu: can limit in BE if we need to

core/src/bundle_stage.rs
```diff
@@ -304,24 +304,25 @@ impl BundleStage {
             &bank_start.working_bank,
         );
 
-        // qos rate-limited a tx in here, drop the bundle
+        // accumulates QoS to metrics
+        qos_service.accumulate_estimated_transaction_costs(
```
buffalu, line 308: i think we might wanna move this to after the `if sanitized_bundle.transactions.len() != num_included`?

buffalu: to where its at right now

jedleggett: you mean inside the if block?  

if you put it after the if block (like it was before) it doesn't get hit for cost model erros

buffalu: hmm i see.. problem here is that it would overcount metrics if the if statement is hit

buffalu: however looking at the metrics for banking stage it looks like it could do something simlar

jedleggett: i want this code to be hit, so i can look for these on bundle stage thread id
https://github.com/jito-foundation/jito-solana/blob/30b47b8ff92081a0b8ee9382ce08baeae4a0cffe/core/src/bundle_stage.rs#L235-L243

buffalu: makes sense... i think banking_stage has the same issue, so probably ok

buffalu: update the estimated: https://github.com/jito-foundation/jito-solana/blob/master/core/src/banking_stage.rs#L1757

here all txs might not be free to lock, which means there might be double counting for estimated cus: https://github.com/jito-foundation/jito-solana/blob/master/core/src/banking_stage.rs#L1774

core/src/bundle_stage.rs
```diff
@@ -1307,16 +1311,27 @@ impl BundleStage {
...
             // leader now, insert new read bundles + as many as can read then return bank
             (Some(bank_start), _) => {
                 consensus_cache_updater.maybe_update(&bank_start.working_bank);
 
+                let is_new_slot = match (last_slot, bundle_stage_leader_stats.current_slot) {
+                    (Some(last_slot), Some(current_slot)) => last_slot != current_slot,
+                    (None, Some(_)) => true,
+                    (_, _) => false,
+                };
+                if is_new_slot && !cost_model_failed_bundles.is_empty() {
```
buffalu, line 1327: can you add a debug log here?

buffalu: like slot number, how many bundles buffered, etc.

buffalu (approved):

buffalu:
lgtm plz squash + merge when ready

buffalu:
have @segfaultdoc take a look too

jedleggett:
gonna update frankfurt-1 and let it bake for a few leader rotations with the log statement, just to verify

jedleggett:
forced pushed to wrong branch

cherry-picking from 1.13.6-jito

PR #226, opened 2023-01-11
Add validator identity pubkey to StakeMeta

steviez:
#### Problem
The vote account associated with a validator is not a permanent link.

#### Summary of Changes
So log the validator identity as well.

Let me know if you all are amenable to pushing this or not; no biggie if you don't want to change it too

### DIFF
--- a/tip-distributor/src/lib.rs
+++ b/tip-distributor/src/lib.rs
@@ -281,6 +281,9 @@ pub struct StakeMeta {
     #[serde(with = "pubkey_string_conversion")]
     pub validator_vote_account: Pubkey,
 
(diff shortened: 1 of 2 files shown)


buffalu:
we're good to pushing this! our CI a bit borked rn with external contributors, but will compile locally

steviez:
> we're good to pushing this! our CI a bit borked rn with external contributors, but will compile locally

Cool, I did run `cargo build` (and the change is pretty trivial as you can see) but no huge rush on pushing this. Also, feel free to use this as a test for sorting out CI with external contributors, I'm not blocked by not having this.

PR #228, opened 2023-01-17
invoke updated init tda instruction

segfaultdoc:
Updates jito-programs submodule.

### DIFF
--- a/validator/src/main.rs
+++ b/validator/src/main.rs
@@ -2653,11 +2653,8 @@ pub fn main() {
     }
     let full_api = matches.is_present("full_rpc_api");
 
-    let identity_keypair = Arc::new(identity_keypair);
-
     let voting_disabled = matches.is_present("no_voting") || restricted_repair_only_mode;
-    let tip_manager_config =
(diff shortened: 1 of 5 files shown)


validator/src/main.rs
```diff
@@ -3567,17 +3562,7 @@ fn tip_manager_config_from_matches(
...
-                    keypair_of(matches, "tip_distribution_account_payer").unwrap_or_else(|| {
-                        if !voting_disabled {
-                            panic!("--tip-distribution-account-payer argument required when validator is voting");
-                        }
-                        Keypair::new()
-                    });
-
-                Arc::new(keypair)
-            },
+            node_identity,
```
buffalu, line 3565: this won't fix our issue, need to use cluster info here + dynamically grab keypair

validator/src/main.rs
```diff
@@ -1831,13 +1831,6 @@ pub fn main() {
                 .takes_value(true)
                 .help("The public key of the tip-distribution program.")
         )
-        .arg(
-            Arg::with_name("tip_distribution_account_payer")
```
buffalu, line 1835: should make sure to add this in release notes

buffalu (approved):

PR #238, opened 2023-01-30
Preallocate Bundle Cost

jedleggett:
#### Problem
Bundles are not landing due to maximum block cost errors.

#### Summary of Changes
Preallocate space for bundles as soon as leader slot starts.

ToDo:

- [x] Release Reserved Space at 80% slot ticks. 
- [ ] initialize Reserved Space with current block limit instead of MAX_BLOCK_UNITS

### DIFF
--- a/core/src/banking_stage.rs
+++ b/core/src/banking_stage.rs
@@ -1749,8 +1749,12 @@ impl BankingStage {
 
         let transaction_costs = qos_service.compute_transaction_costs(txs.iter());
 
-        let (transactions_qos_results, num_included) =
-            qos_service.select_transactions_per_cost(txs.iter(), transaction_costs.iter(), bank);
+        let (transactions_qos_results, num_included) = qos_service.select_transactions_per_cost(
+            txs.iter(),
+            transaction_costs.iter(),
+            bank.slot(),
+            &mut bank.write_cost_tracker().unwrap(),
+        );
 
         let cost_model_throttled_transactions_count = txs.len().saturating_sub(num_included);
 
--- a/core/src/bundle_stage.rs
+++ b/core/src/bundle_stage.rs
@@ -143,7 +143,7 @@ impl BundleReservedSpace {
             .set_block_cost_limit(self.current_tx_block_limit);
 
         debug!(
-            "Slot: {}. Cost Limits Reset. Bundle: {}, TX: {}",
+            "slot: {}. cost limits reset. bundle: {}, txn: {}",
             working_bank.slot(),
             self.current_bundle_block_limit,
             self.current_tx_block_limit,
@@ -171,7 +171,7 @@ impl BundleReservedSpace {
                 .unwrap()
                 .set_block_cost_limit(self.current_tx_block_limit);
             debug!(
-                "Slot: {}. Increased Tx Cost Limit to {}",
+                "slot: {}. increased tx cost limit to {}",
                 working_bank.slot(),
                 self.current_tx_block_limit
             );
@@ -1447,7 +1447,7 @@ impl BundleStage {
                 // Re-Buffer any bundles that didn't fit into last block
                 if !cost_model_failed_bundles.is_empty() {
                     info!(
-                        "Slot {}: Re-buffering {} bundles that failed cost model!",
+                        "slot {}: re-buffering {} bundles that failed cost model.",
                         &bank_start.working_bank.slot(),
                         cost_model_failed_bundles.len()
                     );
--- a/core/src/bundle_stage.rs
+++ b/core/src/bundle_stage.rs
@@ -365,6 +365,7 @@ impl BundleStage {
                 qos_service.compute_transaction_costs(sanitized_bundle.transactions.iter());
 
             let mut cost_tracker = bank_start.working_bank.write_cost_tracker().unwrap();
+
             // Increase block cost limit for bundles
             debug!(
                 "increasing cost limit for bundles: {}",
@@ -1520,7 +1521,7 @@ impl BundleStage {
             current_tx_block_limit: MAX_BLOCK_UNITS.saturating_sub(preallocated_bundle_cost),
             initial_allocated_cost: preallocated_bundle_cost,
             unreserved_ticks: poh_recorder
-                .lock()
+                .write()
                 .unwrap()
                 .ticks_per_slot()
                 .saturating_div(5),
@@ -1528,7 +1529,7 @@ impl BundleStage {
         debug!(
             "initialize bundled reserved space: {preallocated_bundle_cost} cu for {} ticks",
             poh_recorder
-                .lock()
+                .write()
                 .unwrap()
                 .ticks_per_slot()
                 .saturating_sub(reserved_space.unreserved_ticks)
--- a/core/src/bundle_stage.rs
+++ b/core/src/bundle_stage.rs
@@ -35,6 +35,7 @@ use {
             TransactionBalancesSet, TransactionExecutionResult,
         },
         bank_utils,
+        block_cost_limits::MAX_BLOCK_UNITS,
         cost_model::{CostModel, TransactionCost},
         transaction_batch::TransactionBatch,
         vote_sender_types::ReplayVoteSender,
@@ -62,6 +63,7 @@ use {
         thread::{self, Builder, JoinHandle},
         time::{Duration, Instant},
     },
+    uuid::Uuid,
 };
 
 const MAX_BUNDLE_RETRY_DURATION: Duration = Duration::from_millis(10);
@@ -122,6 +124,61 @@ struct AllExecutionResults {
     pub post_balances: (TransactionBalances, TransactionTokenBalances),
 }
 
+struct BundleReservedSpace {
+    current_tx_block_limit: u64,
+    current_bundle_block_limit: u64,
+    initial_allocated_cost: u64,
+    unreserved_ticks: u64,
+}
+
+impl BundleReservedSpace {
+    fn reset_reserved_cost(&mut self, working_bank: &Arc<Bank>) {
+        self.current_tx_block_limit = self
+            .current_bundle_block_limit
+            .saturating_sub(self.initial_allocated_cost);
+
+        working_bank
+            .write_cost_tracker()
+            .unwrap()
+            .set_block_cost_limit(self.current_tx_block_limit);
+
+        debug!(
+            "Slot: {}. Cost Limits Reset. Bundle: {}, TX: {}",
+            working_bank.slot(),
+            self.current_bundle_block_limit,
+            self.current_tx_block_limit,
+        );
+    }
+
+    fn bundle_block_limit(&self) -> u64 {
+        self.current_bundle_block_limit
+    }
+
+    fn tx_block_limit(&self) -> u64 {
+        self.current_tx_block_limit
+    }
+
+    fn update_reserved_cost(&mut self, working_bank: &Arc<Bank>) {
+        if self.current_tx_block_limit != self.current_bundle_block_limit
+            && working_bank
+                .max_tick_height()
+                .saturating_sub(working_bank.tick_height())
+                < self.unreserved_ticks
+        {
+            self.current_tx_block_limit = self.current_bundle_block_limit;
+            working_bank
+                .write_cost_tracker()
+                .unwrap()
+                .set_block_cost_limit(self.current_tx_block_limit);
+            debug!(
+                "Slot: {}. Increased Tx Cost Limit to {}",
+                working_bank.slot(),
+                self.current_tx_block_limit
+            );
+        }
+    }
+}
+
 pub struct BundleStage {
     bundle_thread: JoinHandle<()>,
 }
@@ -140,6 +197,7 @@ impl BundleStage {
         tip_manager: TipManager,
         bundle_account_locker: BundleAccountLocker,
         block_builder_fee_info: &Arc<Mutex<BlockBuilderFeeInfo>>,
+        preallocated_bundle_cost: u64,
     ) -> Self {
         Self::start_bundle_thread(
             cluster_info,
@@ -153,6 +211,7 @@ impl BundleStage {
             bundle_account_locker,
             MAX_BUNDLE_RETRY_DURATION,
             block_builder_fee_info,
+            preallocated_bundle_cost,
         )
     }
 
@@ -169,6 +228,7 @@ impl BundleStage {
         bundle_account_locker: BundleAccountLocker,
         max_bundle_retry_duration: Duration,
         block_builder_fee_info: &Arc<Mutex<BlockBuilderFeeInfo>>,
+        preallocated_bundle_cost: u64,
     ) -> Self {
         const BUNDLE_STAGE_ID: u32 = 10_000;
         let poh_recorder = poh_recorder.clone();
@@ -191,6 +251,7 @@ impl BundleStage {
                     bundle_account_locker,
                     max_bundle_retry_duration,
                     block_builder_fee_info,
+                    preallocated_bundle_cost,
                 );
             })
             .unwrap();
@@ -292,17 +353,41 @@ impl BundleStage {
         bank_start: &BankStart,
         bundle_stage_leader_stats: &mut BundleStageLeaderStats,
         max_bundle_retry_duration: &Duration,
+        reserved_space: &mut BundleReservedSpace,
     ) -> BundleStageResult<()> {
         if sanitized_bundle.transactions.is_empty() {
             return Ok(());
         }
 
-        let tx_costs = qos_service.compute_transaction_costs(sanitized_bundle.transactions.iter());
-        let (transactions_qos_results, num_included) = qos_service.select_transactions_per_cost(
-            sanitized_bundle.transactions.iter(),
-            tx_costs.iter(),
-            &bank_start.working_bank,
-        );
+        // Try to fit bundle into block using modified limits (scoped to guarantee cost_tracker is dropped)
+        let (tx_costs, transactions_qos_results, num_included) = {
+            let tx_costs =
+                qos_service.compute_transaction_costs(sanitized_bundle.transactions.iter());
+
+            let mut cost_tracker = bank_start.working_bank.write_cost_tracker().unwrap();
+            // Increase block cost limit for bundles
+            debug!(
+                "increasing cost limit for bundles: {}",
+                reserved_space.bundle_block_limit()
+            );
+            cost_tracker.set_block_cost_limit(reserved_space.bundle_block_limit());
+            let (transactions_qos_results, num_included) = qos_service
+                .select_transactions_per_cost(
+                    sanitized_bundle.transactions.iter(),
+                    tx_costs.iter(),
+                    bank_start.working_bank.slot(),
+                    &mut cost_tracker,
+                );
+            debug!(
+                "resetting cost limit for normal transactions: {}",
+                reserved_space.tx_block_limit()
+            );
+
+            // Reset block cost limit for normal txs
+            cost_tracker.set_block_cost_limit(reserved_space.tx_block_limit());
+
+            (tx_costs, transactions_qos_results, num_included)
+        };
 
         // accumulates QoS to metrics
         qos_service.accumulate_estimated_transaction_costs(
@@ -312,14 +397,23 @@ impl BundleStage {
             ),
         );
 
-        // either qos rate-limited a tx in here or bundle exceeds max cost, drop the bundle
+        // qos rate-limited a tx in here, drop the bundle
         if sanitized_bundle.transactions.len() != num_included {
             QosService::remove_transaction_costs(
                 tx_costs.iter(),
                 transactions_qos_results.iter(),
                 &bank_start.working_bank,
             );
-            qos_service.report_metrics(bank_start.working_bank.clone());
+            warn!(
+                "bundle dropped, qos rate limit. uuid: {} bundle_cost: {},  block_cost: {}",
+                sanitized_bundle.uuid,
+                tx_costs.iter().map(|c| c.sum()).sum::<u64>(),
+                &bank_start
+                    .working_bank
+                    .read_cost_tracker()
+                    .unwrap()
+                    .block_cost()
+            );
             return Err(BundleExecutionError::ExceedsCostModel);
         }
 
@@ -742,6 +836,7 @@ impl BundleStage {
                 }
             }
         }
+
         Ok(commit_transaction_details)
     }
 
@@ -885,6 +980,7 @@ impl BundleStage {
         last_tip_update_slot: &mut Slot,
         bundle_stage_leader_stats: &mut BundleStageLeaderStats,
         block_builder_fee_info: &Arc<Mutex<BlockBuilderFeeInfo>>,
+        reserved_space: &mut BundleReservedSpace,
     ) {
         let (sanitized_bundles, sanitized_bundle_elapsed) = measure!(
             unprocessed_bundles
@@ -983,7 +1079,8 @@ impl BundleStage {
                 max_bundle_retry_duration,
                 last_tip_update_slot,
                 bundle_stage_leader_stats,
-                block_builder_fee_info
+                block_builder_fee_info,
+                reserved_space,
             ),
             "execute_locked_bundles_elapsed"
         );
@@ -1056,6 +1153,7 @@ impl BundleStage {
         tip_manager: &TipManager,
         max_bundle_retry_duration: &Duration,
         bundle_stage_leader_stats: &mut BundleStageLeaderStats,
+        reserved_space: &mut BundleReservedSpace,
     ) -> BundleStageResult<()> {
         let initialize_tip_accounts_bundle = SanitizedBundle {
             transactions: Self::get_initialize_tip_accounts_transactions(
@@ -1063,6 +1161,7 @@ impl BundleStage {
                 tip_manager,
                 cluster_info,
             )?,
+            uuid: Uuid::default(),
         };
         if !initialize_tip_accounts_bundle.transactions.is_empty() {
             debug!("initialize tip account");
@@ -1079,6 +1178,7 @@ impl BundleStage {
                 bank_start,
                 bundle_stage_leader_stats,
                 max_bundle_retry_duration,
+                reserved_space,
             );
 
             match &result {
@@ -1118,6 +1218,7 @@ impl BundleStage {
         max_bundle_retry_duration: &Duration,
         bundle_stage_leader_stats: &mut BundleStageLeaderStats,
         block_builder_fee_info: &Arc<Mutex<BlockBuilderFeeInfo>>,
+        reserved_space: &mut BundleReservedSpace,
     ) -> BundleStageResult<()> {
         let start_handle_tips = Instant::now();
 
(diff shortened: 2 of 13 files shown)


core/src/bundle_stage.rs
```diff
@@ -292,12 +293,25 @@ impl BundleStage {
...
+        if total_cost <= *reserved_space {
+            *reserved_space = *reserved_space - total_cost;
+            let mut dummy_tx_cost = TransactionCost::new_with_capacity(1);
+            dummy_tx_cost.bpf_execution_cost = total_cost;
+            QosService::remove_transaction_costs(
+                vec![dummy_tx_cost].iter(),
+                vec![Ok(())].iter(),
+                &bank_start.working_bank,
+            );
+        }
```
jedleggett, line 314: There is an edge case, where the OS context switches between these 2 function calls and a tx from a banking stage thread uses the reserved space.  I don't think this will happen very often, but it could be mitigated by pinning this thread to a specific core and reducing it's niceness.

buffalu: at first glance seems fine to me the way it is

core/src/bundle_stage.rs
```diff
@@ -292,12 +293,25 @@ impl BundleStage {
...
         bundle_stage_leader_stats: &mut BundleStageLeaderStats,
         max_bundle_retry_duration: &Duration,
+        reserved_space: &mut u64,
     ) -> BundleStageResult<()> {
         if sanitized_bundle.transactions.is_empty() {
             return Ok(());
         }
 
         let tx_costs = qos_service.compute_transaction_costs(sanitized_bundle.transactions.iter());
+        let total_cost = tx_costs.iter().fold(0, |acc, cost| acc + cost.sum());
```
buffalu, line 303: should be able to do iter().sum() i think?

jedleggett: `error[E0277]: the trait bound `u64: Sum<&TransactionCost>` is not satisfied
    --> core/src/bundle_stage.rs:306:42
     |
306  |         let total_cost = tx_costs.iter().sum();
     |                                          ^^^ the trait `Sum<&TransactionCost>` is not implemented for `u64`
`

buffalu: prob have to give some type hints in the sum stuff

core/src/bundle_stage.rs
```diff
@@ -1307,13 +1346,28 @@ impl BundleStage {
...
+        let last_slot = bundle_stage_leader_stats.current_slot;
         bundle_stage_leader_stats.maybe_report(id, &working_bank_start);
 
         match (working_bank_start, would_be_leader_soon) {
             // leader now, insert new read bundles + as many as can read then return bank
             (Some(bank_start), _) => {
                 consensus_cache_updater.maybe_update(&bank_start.working_bank);
 
+                // If slot changed and currently leader, reserve space
+                if last_slot != bundle_stage_leader_stats.current_slot {
```
buffalu, line 1358: imo it makes sense to move some of this to a special struct, will be easier to grasp

buffalu: all the pre-alloc logic

buffalu: will be easier to add decay in there too

core/src/bundle_stage.rs
```diff
@@ -292,12 +296,34 @@ impl BundleStage {
...
         bundle_stage_leader_stats: &mut BundleStageLeaderStats,
         max_bundle_retry_duration: &Duration,
+        reserved_space: &mut u64,
     ) -> BundleStageResult<()> {
         if sanitized_bundle.transactions.is_empty() {
             return Ok(());
         }
 
         let tx_costs = qos_service.compute_transaction_costs(sanitized_bundle.transactions.iter());
+        let total_cost = tx_costs.iter().map(|c| c.sum()).sum();
```
segfaultdoc, line 306:
```suggestion
        let bundle_cost = tx_costs.iter().map(|c| c.sum()).sum();
```

core/src/bundle_stage.rs
```diff
@@ -292,12 +296,34 @@ impl BundleStage {
...
+        // If there is enough reserved space left, deduct bundle cost from reserved space before adding
+        if total_cost <= *reserved_space {
+            *reserved_space = *reserved_space - total_cost;
```
segfaultdoc, line 309:
```suggestion
            *reserved_space -= total_cost;
```

core/src/bundle_stage.rs
```diff
@@ -292,12 +296,34 @@ impl BundleStage {
...
+            let mut dummy_tx_cost = TransactionCost::new_with_capacity(1);
```
segfaultdoc, line 310: can you rename to `bundle_cost` or something along those lines?

core/src/bundle_stage.rs
```diff
@@ -292,12 +296,34 @@ impl BundleStage {
...
+                vec![Ok(())].iter(),
+                &bank_start.working_bank,
+            );
+            info!(
+                "Bundle uses {} of reserved space. Remaining: {}",
+                total_cost, *reserved_space
+            );
+        } else {
+            info!(
+                "Bundle Cost {} exceeds available reserved space {}",
```
segfaultdoc, line 323:
```suggestion
                "Bundle Cost {total_cost} exceeds available reserved space {reserved_space}",
```

core/src/bundle_stage.rs
```diff
@@ -292,12 +296,34 @@ impl BundleStage {
...
+            *reserved_space = *reserved_space - total_cost;
+            let mut dummy_tx_cost = TransactionCost::new_with_capacity(1);
+            dummy_tx_cost.bpf_execution_cost = total_cost;
+            QosService::remove_transaction_costs(
+                vec![dummy_tx_cost].iter(),
+                vec![Ok(())].iter(),
+                &bank_start.working_bank,
+            );
+            info!(
+                "Bundle uses {} of reserved space. Remaining: {}",
```
segfaultdoc, line 318:
```suggestion
                "Bundle uses {total_cost} of reserved space. Remaining: {reserved_space}",
```

core/src/bundle_stage.rs
```diff
@@ -1307,13 +1349,46 @@ impl BundleStage {
...
+                let is_new_slot = match (last_slot, bundle_stage_leader_stats.current_slot) {
+                    (Some(last_slot), Some(current_slot)) => last_slot != current_slot,
+                    (None, Some(_)) => true,
+                    (_, _) => false,
+                };
+                if is_new_slot {
+                    // Reserve space from cost tracker
+                    // ToDo: Figure out why this not working
+                    // *reserved_space = preallocated_bundle_cost;
+                    *reserved_space = 1000000;
```
segfaultdoc, line 1370: can you parameterize this?

core/src/bundle_stage.rs
```diff
@@ -1307,13 +1349,46 @@ impl BundleStage {
...
+        let last_slot = bundle_stage_leader_stats.current_slot;
         bundle_stage_leader_stats.maybe_report(id, &working_bank_start);
 
         match (working_bank_start, would_be_leader_soon) {
             // leader now, insert new read bundles + as many as can read then return bank
             (Some(bank_start), _) => {
                 consensus_cache_updater.maybe_update(&bank_start.working_bank);
 
+                // If slot changed and currently leader, reserve space
+                let is_new_slot = match (last_slot, bundle_stage_leader_stats.current_slot) {
```
segfaultdoc, line 1361: I think we'll want a better way to be notified of new slots. Right now this only gets triggered when we receive bundles, but banking_stage could be well into using up all the CUs by the time this receives a bundle and reserves space

core/src/bundle_stage.rs
segfaultdoc: nvm see you have it commented it out

core/src/bundle_stage.rs
jedleggett: this get's called every SLOT_BOUNDARY_CHECK_PERIOD whether there are bundles to process or not
https://github.com/jito-foundation/jito-solana/blob/ce881f5413cb319d8e55fc3b44d01bcb0289e38c/core/src/bundle_stage.rs#L1402-L1408

it's kinda weird flow, but i was trying to make minimal changes to what's there

validator/src/main.rs
```diff
@@ -114,6 +114,7 @@ const DEFAULT_MIN_SNAPSHOT_DOWNLOAD_SPEED: u64 = 10485760;
 // The maximum times of snapshot download abort and retry
 const MAX_SNAPSHOT_DOWNLOAD_ABORT: u32 = 5;
 const MILLIS_PER_SECOND: u64 = 1000;
+const DEFAULT_PREALLOCATED_BUNDLE_COST: u64 = 1000000;
```
buffalu, line 117: arbs are ~150k compute units, so i think we should set default higher to like 2M or 3M

jedleggett: make sense

validator/src/main.rs
```diff
@@ -1845,6 +1847,14 @@ pub fn main() {
                 .takes_value(true)
                 .help("The commission validator takes from tips expressed in basis points.")
         )
+        .arg(
+            Arg::with_name("preallocated_bundle_cost")
```
buffalu, line 1851: i don't see this being used anywhere, maybe not working bc its not piped through to config?

jedleggett: that was it, fixed now

core/src/bundle_stage.rs
jedleggett: currently, every 10 ms

core/src/bundle_stage.rs
```diff
@@ -122,6 +122,80 @@ struct AllExecutionResults {
...
+        );
+    }
+
+    fn use_reserved_cost(&mut self, bundle_cost: u64, working_bank: &Arc<Bank>) {
+        // If there is enough reserved space left, deduct bundle cost from reserved space before adding
+        if bundle_cost <= self.current_allocated_cost {
+            self.current_allocated_cost -= bundle_cost;
+            let mut dummy_tx_cost = TransactionCost::new_with_capacity(1);
+            dummy_tx_cost.bpf_execution_cost = bundle_cost;
+            QosService::remove_transaction_costs(
```
segfaultdoc, line 161: Should this be calling `update_transaction_costs`?

core/src/bundle_stage.rs
```diff
@@ -122,6 +122,80 @@ struct AllExecutionResults {
...
+        }
+    }
+
+    fn update_reserved_cost(&mut self, working_bank: &Arc<Bank>) {
+        let target_reserved_cost = self.initial_allocated_cost * working_bank.tick_height()
+            / working_bank.ticks_per_slot();
+        if self.current_allocated_cost > target_reserved_cost {
+            let mut dummy_tx_cost = TransactionCost::new_with_capacity(1);
+            dummy_tx_cost.bpf_execution_cost = self.current_allocated_cost - target_reserved_cost;
+            QosService::remove_transaction_costs(
```
segfaultdoc, line 184: ditto here should we be calling `update_or_remove_transaction_costs` and have it hit the code path that calls into `update_transaction_costs`?

core/src/bundle_stage.rs
```diff
@@ -292,12 +370,16 @@ impl BundleStage {
...
+        reserved_space: &mut BundleReservedSpace,
     ) -> BundleStageResult<()> {
         if sanitized_bundle.transactions.is_empty() {
             return Ok(());
         }
 
         let tx_costs = qos_service.compute_transaction_costs(sanitized_bundle.transactions.iter());
+        let bundle_cost = tx_costs.iter().map(|c| c.sum()).sum();
+        // Use available reserved space
+        reserved_space.use_reserved_cost(bundle_cost, &bank_start.working_bank);
```
segfaultdoc, line 382: Should early return if unable to reserve space

core/src/bundle_stage.rs
segfaultdoc: hm I think I see what you're doing:
1. on new slot "reserve" max allowed cost for bundles
2. on receiving a new bundle you free up space for that bundle
3. then try and allocate it again thru `select_transactions_per_cost`

segfaultdoc: still grokking but if I'm understanding correctly, don't we run the risk of banking stage eating up the freed space b/w steps 2 and 3?

core/src/bundle_stage.rs
jedleggett: that would just require an extra definition of an estimated tx cost, a comparison, then passing the difference into cost_tracker.sub_transaction_execution_cost

remove_transaction_cost results in identical behavior from the cost_tracker perspective

core/src/bundle_stage.rs
jedleggett: fixed by holding cost_tracker lock

core/src/bundle_stage.rs
```diff
@@ -122,6 +123,53 @@ struct AllExecutionResults {
...
+struct BundleReservedSpace {
+    current_tx_block_limit: u64,
+    current_bundle_block_limit: u64,
+    initial_allocated_cost: u64,
+}
+
+impl BundleReservedSpace {
+    fn reset_reserved_cost(&mut self, working_bank: &Arc<Bank>) {
+        self.current_bundle_block_limit = MAX_BLOCK_UNITS;
+        self.current_tx_block_limit = MAX_BLOCK_UNITS - self.initial_allocated_cost;
```
buffalu, line 135: use saturating math here and elsewhere. master doesn't build right now but when it does it'll compain about that

core/src/bundle_stage.rs
```diff
@@ -1399,6 +1493,12 @@ impl BundleStage {
 
         let mut unprocessed_bundles: VecDeque<PacketBundle> = VecDeque::with_capacity(1000);
         let mut cost_model_failed_bundles: VecDeque<PacketBundle> = VecDeque::with_capacity(1000);
+        let mut reserved_space = BundleReservedSpace {
+            current_bundle_block_limit: MAX_BLOCK_UNITS,
+            current_tx_block_limit: MAX_BLOCK_UNITS - preallocated_bundle_cost,
```
buffalu, line 1498: saturating math

core/src/bundle_stage.rs
```diff
@@ -122,6 +123,53 @@ struct AllExecutionResults {
...
+    initial_allocated_cost: u64,
+}
+
+impl BundleReservedSpace {
+    fn reset_reserved_cost(&mut self, working_bank: &Arc<Bank>) {
+        self.current_bundle_block_limit = MAX_BLOCK_UNITS;
+        self.current_tx_block_limit = MAX_BLOCK_UNITS - self.initial_allocated_cost;
+
+        working_bank.write_cost_tracker().unwrap().set_limits(
+            MAX_WRITABLE_ACCOUNT_UNITS,
```
buffalu, line 138: these might mess with tests or have other unexpected behavior. can we set limit to current limit for values that we don't wanna change?

buffalu: seems like these are internal values, maybe worth exposing as read-only accessors?

core/src/bundle_stage.rs
```diff
@@ -122,6 +123,53 @@ struct AllExecutionResults {
...
 }
 
+struct BundleReservedSpace {
+    current_tx_block_limit: u64,
+    current_bundle_block_limit: u64,
+    initial_allocated_cost: u64,
+}
+
+impl BundleReservedSpace {
+    fn reset_reserved_cost(&mut self, working_bank: &Arc<Bank>) {
```
buffalu, line 133: if i understand correctly, instead of reserving space we're artificially limiting space?

core/src/bundle_stage.rs
```diff
@@ -122,6 +123,53 @@ struct AllExecutionResults {
...
+        self.current_bundle_block_limit
+    }
+
+    fn tx_block_limit(&self) -> u64 {
+        self.current_tx_block_limit
+    }
+
+    fn update_reserved_cost(&mut self, working_bank: &Arc<Bank>) {
+        if self.current_tx_block_limit != self.current_bundle_block_limit
+            && working_bank.tick_height() > working_bank.ticks_per_slot() * 4 / 5
```
buffalu, line 161: nit: saturating math

buffalu: also nit: ticks_per_slot shouldn't change so could set this on init

core/src/bundle_stage.rs
```diff
@@ -122,6 +123,53 @@ struct AllExecutionResults {
...
+        {
+            info!(
```
buffalu, line 163: we missing call to set_limits here?

core/src/bundle_stage.rs
```diff
@@ -292,16 +344,32 @@ impl BundleStage {
...
+        reserved_space: &mut BundleReservedSpace,
     ) -> BundleStageResult<()> {
         if sanitized_bundle.transactions.is_empty() {
             return Ok(());
         }
 
         let tx_costs = qos_service.compute_transaction_costs(sanitized_bundle.transactions.iter());
+        let bundle_cost: u64 = tx_costs.iter().map(|c| c.sum()).sum();
+        let cost_tracker = &mut bank_start.working_bank.write_cost_tracker().unwrap();
+        // Increase blcok cost limit for bundles
```
buffalu, line 356: typo

core/src/bundle_stage.rs
buffalu: oh nvmd, can you add comment in here on why we don't do it? seems like we need to "atomically swap" compute units or something

core/src/bundle_stage.rs
```diff
@@ -292,16 +344,32 @@ impl BundleStage {
...
         bundle_stage_leader_stats: &mut BundleStageLeaderStats,
         max_bundle_retry_duration: &Duration,
+        reserved_space: &mut BundleReservedSpace,
     ) -> BundleStageResult<()> {
         if sanitized_bundle.transactions.is_empty() {
             return Ok(());
         }
 
         let tx_costs = qos_service.compute_transaction_costs(sanitized_bundle.transactions.iter());
+        let bundle_cost: u64 = tx_costs.iter().map(|c| c.sum()).sum();
```
buffalu, line 354: ultra nit: move this down to log line where its used

core/src/bundle_stage.rs
```diff
@@ -292,16 +344,32 @@ impl BundleStage {
...
+        let cost_tracker = &mut bank_start.working_bank.write_cost_tracker().unwrap();
+        // Increase blcok cost limit for bundles
+        cost_tracker.set_limits(
```
buffalu, line 357: im a little lost on this stuff, can you explain more?

buffalu: are you essentially increasing block limit, adding these txs, then decreasing it again to fit in?

core/src/bundle_stage.rs
```diff
@@ -312,14 +380,21 @@ impl BundleStage {
...
+        // qos rate-limited a tx in here, drop the bundle
         if sanitized_bundle.transactions.len() != num_included {
             QosService::remove_transaction_costs(
                 tx_costs.iter(),
                 transactions_qos_results.iter(),
                 &bank_start.working_bank,
             );
-            qos_service.report_metrics(bank_start.working_bank.clone());
+            warn!(
+                "Bundle dropped, QoS rate limit!!  Bundle Cost {bundle_cost},  Block Cost {}",
```
buffalu, line 391: can you add bundle uuid?

buffalu: also nit: all lower case, remove exclamations, etc. to match log format

buffalu: bundle_cost: {}, block_cost: {}

core/src/bundle_stage.rs
buffalu: so at worse, this will kick in 80% through slot + 10ms -> 400 * 0.8 =330ms through slot

core/src/bundle_stage.rs
jedleggett: i think it does need to be called here, else it wouldn't actually go into effect until a bundle was processed

buffalu: it seems like it’d be applied later on slot boundary check with working bank?

buffalu: oh nvmd i think we need it here. so in that case, for last 70-80ms of block, compute limit back to 48m units. if we get bundle in that time, maybe fits maybe doesn't

core/src/bundle_stage.rs
jedleggett: wdyt about making the inputs Option<u64> and just ignore if None?

i just added the read-only accessors, and it makes the call to set_limits a lot longer

buffalu: https://github.com/jito-foundation/jito-solana/blob/master/core/src/bundle_stage.rs#L1608

do we handle this correctly

jedleggett: nvm, that function is called all over the place

i think i'll just make a function that only sets the block cost limit

core/src/bundle_stage.rs
jedleggett: right, and without this call, if we don't get a bundle, then it never resets to 48M

core/src/bundle_stage.rs
jedleggett: made another update to fix tests

i only reference MAX_BLOCK_UNITS in init now, and preserve previous limit in update

core/src/banking_stage.rs
```diff
@@ -1749,8 +1749,12 @@ impl BankingStage {
 
         let transaction_costs = qos_service.compute_transaction_costs(txs.iter());
 
-        let (transactions_qos_results, num_included) =
-            qos_service.select_transactions_per_cost(txs.iter(), transaction_costs.iter(), bank);
+        let (transactions_qos_results, num_included) = qos_service.select_transactions_per_cost(
+            txs.iter(),
+            transaction_costs.iter(),
+            bank.slot(),
+            &mut bank.write_cost_tracker().unwrap(),
```
buffalu, line 1756: when does this get dropped?

core/src/qos_service.rs
```diff
@@ -126,22 +126,22 @@ impl QosService {
         &self,
         transactions: impl Iterator<Item = &'a SanitizedTransaction>,
         transactions_costs: impl Iterator<Item = &'a TransactionCost>,
-        bank: &Arc<Bank>,
+        slot: Slot,
+        cost_tracker: &mut RwLockWriteGuard<CostTracker>,
```
buffalu, line 130: could you pass in &mut CostTracker and have this still work?

buffalu: idk if method needs anything related to RwLockWriteGuard

core/src/bundle_stage.rs
```diff
@@ -1399,6 +1514,26 @@ impl BundleStage {
 
         let mut unprocessed_bundles: VecDeque<PacketBundle> = VecDeque::with_capacity(1000);
         let mut cost_model_failed_bundles: VecDeque<PacketBundle> = VecDeque::with_capacity(1000);
+        // Initialize block limits and open up last 20% of ticks to non-bundle transactions
+        let mut reserved_space = BundleReservedSpace {
+            current_bundle_block_limit: MAX_BLOCK_UNITS,
```
buffalu, line 1519: can you use bank current block limit here and below? i think might break tests that force a certain block limit size?

buffalu: can prob get a bank from poh_recorder

jedleggett: i initialize with appropriate limits in the tests below, none of the tests currently reference this function

but, i agree, it would be better to use current block limit, i just couldn't convince myself that there would be a bank available when this line hits

maybe i just need to wait here until a bank is availabe?

core/src/qos_service.rs
jedleggett: i guess i didn't quite understand how the RwLock works

my goal here is to be able to hold the lock on the cost_tracker over multiple function calls

are you saying if i pass this into the function it will still hold the write lock as intended?
`let mut cost_tracker = bank_start.working_bank.write_cost_tracker().unwrap().deref_mut()`

core/src/bundle_stage.rs
buffalu: let's just add a TODO and figure it out later, wanna get this in and running on our machines to make sure its good for release

core/src/qos_service.rs
buffalu: talked offline, should be able to take a `&mut CostTracker` and not have this method worry about whether its a mutex, rwlock, or CostTracker w/ no synchronization

core/src/bundle_stage.rs
jedleggett: i checked the logs on this, and there's ~10 minutes between when this currently gets initialized and when the first bank arrives

i don't see any reason why waiting would hurt, but i'd like to think it throught some more

buffalu: ok sg

buffalu (approved):

core/src/bundle_stage.rs
```diff
@@ -1399,6 +1515,26 @@ impl BundleStage {
...
+            current_tx_block_limit: MAX_BLOCK_UNITS.saturating_sub(preallocated_bundle_cost),
+            initial_allocated_cost: preallocated_bundle_cost,
+            unreserved_ticks: poh_recorder
+                .write()
+                .unwrap()
+                .ticks_per_slot()
+                .saturating_div(5),
```
esemeniuc, line 1527: curious why div by 5?

esemeniuc: maybe leave comment

core/src/bundle_stage.rs
```diff
@@ -312,14 +398,23 @@ impl BundleStage {
...
+        // qos rate-limited a tx in here, drop the bundle
         if sanitized_bundle.transactions.len() != num_included {
             QosService::remove_transaction_costs(
                 tx_costs.iter(),
                 transactions_qos_results.iter(),
                 &bank_start.working_bank,
             );
-            qos_service.report_metrics(bank_start.working_bank.clone());
+            warn!(
+                "bundle dropped, qos rate limit. uuid: {} bundle_cost: {},  block_cost: {}",
```
esemeniuc, line 409: extra space before block_cost

core/src/bundle_stage.rs
```diff
@@ -312,14 +398,23 @@ impl BundleStage {
...
+                sanitized_bundle.uuid,
+                tx_costs.iter().map(|c| c.sum()).sum::<u64>(),
```
esemeniuc, line 411:
```suggestion
                tx_costs.iter().map(TransactionCost::sum).sum::<u64>(),
```

core/src/bundle_stage.rs
jedleggett: it's the 20% comment above

core/src/bundle_stage.rs
jedleggett: this is printing current block cost when bundle wouldn't fit

independent of extra space or not

core/src/bundle_stage.rs
jedleggett: ahh, that is better, but i don't want to change without testing, and i don't want to bring down all of our validators to test a style change

PR #240, opened 2023-02-01
bugfix: conditionally compile with debug flags

segfaultdoc:


dev/Dockerfile
```diff
@@ -33,9 +33,15 @@ RUN mkdir -p docker-output
...
+ARG debug
+
 # Uses docker buildkit to cache the image.
 # /usr/local/cargo/git needed for crossbeam patch
 RUN --mount=type=cache,mode=0777,target=/solana/target \
     --mount=type=cache,mode=0777,target=/usr/local/cargo/registry \
     --mount=type=cache,mode=0777,target=/usr/local/cargo/git \
-    RUSTFLAGS='-g -C force-frame-pointers=yes' ./cargo stable build --release && cp target/release/solana* ./docker-output
+    if [ "$debug" = "false" ] ; then \
+      ./cargo stable build --release && cp target/release/solana* ./docker-output && echo "compiling without debug flags......"; \
```
buffalu, line 44: nit: printed after it runs. can prob just remove the echo

f
```diff
@@ -9,7 +9,10 @@ GIT_SHA="$(git describe --always --dirty)"
 
 echo $GIT_SHA
 
+DEBUG_FLAGS=${1-false}
```
buffalu, line 12: can you put example here on how to enable debug

PR #244, opened 2023-02-08
Jl/buffer bundles 1.13.6 jito

jedleggett:
#### Problem
Bundles are failing the QOS cost model when blocks are full.  

#### Summary of Changes
Buffer bundles that exceed the QOS cost model.


segfaultdoc:
can u update this to include bugfix?

PR #246, opened 2023-02-08
fix smol bug

segfaultdoc:


esemeniuc (approved):

buffalu:
imo hold off on this, if not fixing bug we shouldn't merge code

PR #247, opened 2023-02-09
Account for block builder fee when generating excess tip balance

buffalu:
#### Problem
The block builder fee wasn't taken into account when doing mev claims. This was resulting in the mev claim process thinking it had more lamports than it did, causing the claim process to distribute too much to entities and the last jito-solana leader in an epoch having a failed claim on the last claim account.

#### Summary of Changes
Add more stringent checks around accounts existing.
Take into account block builder fee

### DIFF
--- a/tip-distributor/src/stake_meta_generator_workflow.rs
+++ b/tip-distributor/src/stake_meta_generator_workflow.rs
@@ -165,24 +165,35 @@ pub fn generate_stake_meta_collection(
     // We assume that the rewards sitting in the tip program PDAs are cranked out by the time all of
     // the rewards are claimed.
     let tip_accounts = derive_tip_payment_pubkeys(tip_payment_program_id);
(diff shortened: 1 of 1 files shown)


tip-distributor/src/stake_meta_generator_workflow.rs
```diff
@@ -165,24 +165,36 @@ pub fn generate_stake_meta_collection(
...
+            tip_account
+                .lamports()
+                .checked_sub(bank.get_minimum_balance_for_rent_exemption(tip_account.data().len()))
+                .expect("tip balance underflow")
         })
         .sum();
+    // matches math in tip payment program
+    let block_builder_tips = excess_tip_balances
+        .checked_mul(bb_commission_pct)
+        .expect("bb_commission overflow")
```
esemeniuc, line 192:
```suggestion
        .expect("block_builder_tips overflow")
```

esemeniuc (approved):

PR #221, opened 2022-12-23
fix: simulate bundle rpc was returning sucess on empty list

anoushk1234:
#### Problem
The function simulate_bundle in rpc.rs would return success even if list of encoded transactions was empty, this is misleading.

#### Summary of Changes
- Changed simulate_bundle to early return error if rpc_bundle_request has empty list of encoded transactions
- Added test case test_rpc_simulate_bundle_error_on_empty_list that verifies that the function returns the InvalidParams error if an empty list is passed

Fixes #165

PR #257, opened 2023-02-14
fix: simulate bundle rpc was returning sucess on empty list

segfaultdoc:
hacky way around gated buildkite, copy of [this](https://github.com/jito-foundation/jito-solana/pull/221)


rpc/src/rpc.rs
```diff
@@ -9116,4 +9122,75 @@ pub mod tests {
...
+                         {{ "encoding": "base64", "addresses": [] }}
+                      ],
+                      "postExecutionAccountsConfigs": [
+                         {{ "encoding": "base64", "addresses": [] }},
+                         {{ "encoding": "base64", "addresses": ["{}"] }}
+                      ]
+                    }}
+                 ]
+             }}"#,
+            skip_sig_verify,
```
jedleggett, line 9178: nit: doesn't `format! `accept these inline now?

jedleggett (approved):
lgtm

PR #260, opened 2023-02-15
fix smol bug

segfaultdoc:
Fixes bug where bundles buffered due to cost model failures are drained if current validator not a leader soon or now.

### DIFF
--- a/core/src/bundle_stage.rs
+++ b/core/src/bundle_stage.rs
@@ -1364,6 +1364,10 @@ impl BundleStage {
             }
             _ => {}
         }
+
+        if !would_be_leader_soon {
+            cost_model_failed_bundles.clear();
+        }
     }
 
     #[allow(clippy::too_many_arguments)]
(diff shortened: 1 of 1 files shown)


buffalu:
can you add description for future on bug fixed?

core/src/bundle_stage.rs
```diff
@@ -1315,54 +1315,53 @@ impl BundleStage {
...
-                if is_new_slot && !cost_model_failed_bundles.is_empty() {
-                    debug!(
-                        "Slot {}: Re-buffering {} bundles that failed cost model!",
-                        &bank_start.working_bank.slot(),
-                        cost_model_failed_bundles.len()
-                    );
-                    unprocessed_bundles.extend(cost_model_failed_bundles.drain(..));
-                }
+        if !would_be_leader_soon {
+            saturating_add_assign!(
```
jedleggett, line 1319: maybe add
`unprocessed_bundles.extend(cost_model_failed_bundles.drain(..))`
here

PR #261, opened 2023-02-20
Backport #240 on 1.14

buffalu:
Backports #240 onto v1.14.
Adds use of ./cargo instead of normal cargo


f
```diff
@@ -9,7 +11,10 @@ GIT_SHA="$(git describe --always --dirty)"
 
 echo $GIT_SHA
 
+DEBUG_FLAGS=${1-false}
```
esemeniuc, line 14: missing colon? https://stackoverflow.com/a/2013589/3408577

esemeniuc (approved):

PR #276, opened 2023-03-16
Seg/v1.14

segfaultdoc:
Backports Dockerfile from master


esemeniuc (approved):

buffalu:
breaks this https://github.com/jito-foundation/jito-solana/blob/master/version/src/lib.rs#L70

PR #281, opened 2023-03-20
Fix Tests

jedleggett:
#### Problem
Locking Working Bank and PoH Recorder was causing deadlock during tests.

#### Summary of Changes
Fetch values needed from Bank and PoH early.


esemeniuc (approved):

core/src/bundle_stage.rs
```diff
@@ -1498,6 +1498,7 @@ impl BundleStage {
     ) {
         const LOOP_STATS_METRICS_PERIOD: Duration = Duration::from_secs(1);
 
+        let ticks_per_slot = poh_recorder.read().unwrap().ticks_per_slot();
```
jedleggett, line 1501: this was not breaking the test, but i went ahead and copied the pattern that fixed the test

buffalu (approved):

PR #290, opened 2023-03-29
Identity Hotswap

jedleggett:
#### Problem
Using the Admin RPC, it is possible to change the identity of a running validator without restart (hotswap).  The Jito back end connections (Block Engine and Relayer) have no way to detect this change, and when swapping from a staked identity to an unstaked identity, the connection to the back end persists and auth continues to retry with the old (staked) pubkey.  When the hot spare is switch on (unstaked -> staked) there are 2 running validators trying to connect to the backend continuously overwriting each other's connections.

#### Summary of Changes
Add logic to detect identity change inside the back end proxies (relayer and block engine).  When detected, exit the authentication loop, emit a metric, and start over with new identity.


validator/src/admin_rpc_service.rs
```diff
@@ -33,6 +33,9 @@ pub struct AdminRpcRequestMetadataPostInit {
     pub cluster_info: Arc<ClusterInfo>,
     pub bank_forks: Arc<RwLock<BankForks>>,
     pub vote_account: Pubkey,
+    // ToDo (JL):
+    // pub maybe_block_engine_config: Arc<RwLock<Option<BlockEngineConfig>>>,
```
segfaultdoc, line 37: rm?

validator/src/admin_rpc_service.rs
```diff
@@ -183,6 +186,23 @@ pub trait AdminRpc {
 
     #[rpc(meta, name = "contactInfo")]
     fn contact_info(&self, meta: Self::Metadata) -> Result<AdminRpcContactInfo>;
+
+    // ToDo (JL):
+    // #[rpc(meta, name = "setBlockEngineConfig")]
```
segfaultdoc, line 191: rm?

validator/src/admin_rpc_service.rs
```diff
@@ -324,6 +344,28 @@ impl AdminRpc for AdminRpcImpl {
     fn contact_info(&self, meta: Self::Metadata) -> Result<AdminRpcContactInfo> {
         meta.with_post_init(|post_init| Ok(post_init.cluster_info.my_contact_info().into()))
     }
+
+    // fn set_block_engine_config(
```
segfaultdoc, line 348: rm

validator/src/main.rs
```diff
@@ -3334,11 +3334,15 @@ pub fn main() {
         error!("Failed to start validator: {:?}", e);
         exit(1);
     });
+
     *admin_service_post_init.write().unwrap() =
         Some(admin_rpc_service::AdminRpcRequestMetadataPostInit {
             bank_forks: validator.bank_forks.clone(),
             cluster_info: validator.cluster_info.clone(),
             vote_account,
+            // ToDo (JL):
```
segfaultdoc, line 3343: rm

core/src/proxy/auth.rs
```diff
@@ -42,206 +40,160 @@ impl Interceptor for AuthInterceptor {
...
-                }
-                // Sleep and do nothing if neither token is close to expired,
-                (false, false) => sleep(sleep_interval).await,
-            }
-        }
-
-        Ok(())
-    }
+/// Generates an auth challenge then generates and returns validated auth tokens.
+pub async fn generate_auth_tokens(
```
segfaultdoc, line 44: did this code change?

core/src/proxy/auth.rs
```diff
@@ -42,206 +40,160 @@ impl Interceptor for AuthInterceptor {
...
+
+    if should_generate_new_tokens {
+        let kp = cluster_info.keypair().clone();
+
+        let (new_access_token, new_refresh_token) = timeout(
+            *connection_timeout,
+            generate_auth_tokens(auth_service_client, kp.as_ref()),
+        )
+        .await
+        .map_err(|_| ProxyError::MethodTimeout("refresh_access_token".to_string()))?
```
segfaultdoc, line 136:
```suggestion
        .map_err(|_| ProxyError::MethodTimeout("generate_auth_tokens".to_string()))?
```

core/src/proxy/auth.rs
```diff
@@ -42,206 +40,160 @@ impl Interceptor for AuthInterceptor {
...
+        .map_err(|e| ProxyError::MethodError(e.to_string()))?;
+
+        *access_token.lock().unwrap() = new_access_token.clone();
+        *refresh_token = new_refresh_token.clone();
+
+        *num_full_refreshes += 1;
+        datapoint_info!(
+            "auth_tokens_update_loop-tokens_generated",
```
segfaultdoc, line 144: let's pass this metric name down from above and have it be prepended with relayer or block-engine depending on what stage is invoking

core/src/proxy/auth.rs
jedleggett: just moved out of module, changed error type, and added debug statements

validator/src/admin_rpc_service.rs
jedleggett: yeah, for next pr

core/src/proxy/auth.rs
jedleggett: in that case, do we need the url?

core/src/proxy/auth.rs
```diff
@@ -42,206 +40,162 @@ impl Interceptor for AuthInterceptor {
...
+    datapoint_access_refresh: &'static str,
+    auth_uri: &String,
+    access_token: &Arc<Mutex<Token>>,
+    refresh_token: &mut Token,
+    cluster_info: &Arc<ClusterInfo>,
+    connection_timeout: &Duration,
+    auth_refresh_lookahead: u64,
+    num_full_refreshes: &mut u64,
+    num_refresh_access_token: &mut u64,
+) -> crate::proxy::Result<()> {
```
buffalu, line 109: is it possible to return (Option<Token>, Option<Token>) instead of passing in mutable access_token + refresh_token and metrics? pull metrics tracking to caller depending on what's returned

core/src/proxy/block_engine_stage.rs
```diff
@@ -300,10 +369,36 @@ impl BlockEngineStage {
                     block_engine_stats = BlockEngineStageStats::default();
                 }
                 _ = maintenance_tick.tick() => {
-                    let block_builder_info = client.get_block_builder_fee_info(BlockBuilderFeeInfoRequest{}).await?.into_inner();
+
+                    if cluster_info.id() != keypair.pubkey() {
```
buffalu, line 373: imo we should generate a signal to catch this faster. need to provide a signal mechanism from set_identity

core/src/proxy/block_engine_stage.rs
```diff
@@ -272,25 +327,39 @@ impl BlockEngineStage {
...
+        refresh_token: &mut Token,
+        keypair: Arc<Keypair>,
+        cluster_info: &Arc<ClusterInfo>,
+        connection_timeout: &Duration,
     ) -> crate::proxy::Result<()> {
         const METRICS_TICK: Duration = Duration::from_secs(1);
         const MAINTENANCE_TICK: Duration = Duration::from_secs(10 * 60);
-
+        // Lookahead by Maintenance Tick plus 25%
+        const AUTH_REFRESH_LOOKAHEAD: u64 = MAINTENANCE_TICK
```
buffalu, line 344: lookahead kinda confusing, would make more sense if REFRESH_WITHIN_S or something like that saying it'll refresh if the tokens expire within N seconds

core/src/proxy/block_engine_stage.rs
```diff
@@ -272,25 +327,39 @@ impl BlockEngineStage {
...
+        access_token: Arc<Mutex<Token>>,
+        refresh_token: &mut Token,
+        keypair: Arc<Keypair>,
+        cluster_info: &Arc<ClusterInfo>,
+        connection_timeout: &Duration,
     ) -> crate::proxy::Result<()> {
         const METRICS_TICK: Duration = Duration::from_secs(1);
         const MAINTENANCE_TICK: Duration = Duration::from_secs(10 * 60);
-
+        // Lookahead by Maintenance Tick plus 25%
```
buffalu, line 343: unnecessary comment. what is obvious, why isn't obvious.

why = need to ensure AUTH_REFRESH_LOOKAHEAD >= MAINTENANCE_TICK. if AUTH_REFRESH_LOOKAHEAD < MAINTENANCE_TICK, then you might end up with MAINTENANCE_TICK - AUTH_REFRESH_LOOKAHEAD seconds where the access token is invalid

core/src/proxy/block_engine_stage.rs
```diff
@@ -272,25 +327,39 @@ impl BlockEngineStage {
...
         block_builder_fee_info: &Arc<Mutex<BlockBuilderFeeInfo>>,
+        mut auth_client: AuthServiceClient<Channel>,
+        access_token: Arc<Mutex<Token>>,
+        refresh_token: &mut Token,
+        keypair: Arc<Keypair>,
+        cluster_info: &Arc<ClusterInfo>,
+        connection_timeout: &Duration,
     ) -> crate::proxy::Result<()> {
         const METRICS_TICK: Duration = Duration::from_secs(1);
         const MAINTENANCE_TICK: Duration = Duration::from_secs(10 * 60);
```
buffalu, line 342: is there a way to do this more dynamically but also simple (simple is key)? if tokens are good for 5 minutes, this logic won't work bc needs to be refreshed more frequently than MAINTENANCE_TICK

buffalu: it might make more sense to do re-auth in the 1s check

core/src/proxy/auth.rs
buffalu: mutating a bunch of state through parameters feels kinda icky

core/src/proxy/relayer_stage.rs
```diff
@@ -147,89 +134,136 @@ impl RelayerStage {
...
+                &heartbeat_tx,
+                &packet_tx,
+                &verified_packet_tx,
+                &exit,
+                &CONNECTION_TIMEOUT,
+            )
+            .await
+            {
+                Ok(_) => {
+                    backoff_sec = 0;
```
buffalu, line 162: i don't think this will ever return ok? im okay removing backoff_sec + waiting some small amount of time between reconnects (1s, 2s, or 5s)

core/src/proxy/block_engine_stage.rs
buffalu: less hacky would be doing the check once per second

core/src/proxy/relayer_stage.rs
```diff
@@ -278,25 +316,43 @@ impl RelayerStage {
...
+        mut auth_client: AuthServiceClient<Channel>,
+        access_token: Arc<Mutex<Token>>,
+        refresh_token: &mut Token,
+        keypair: Arc<Keypair>,
+        cluster_info: &Arc<ClusterInfo>,
+        connection_timeout: &Duration,
     ) -> crate::proxy::Result<()> {
         const METRICS_TICK: Duration = Duration::from_secs(1);
+        const MAINTENANCE_TICK: Duration = Duration::from_secs(10 * 60);
+        // Lookahead by Maintenance Tick plus 25%
```
buffalu, line 331: same comment here on "why", not "what"

core/src/proxy/block_engine_stage.rs
jedleggett: I moved to metrics_tick.  Honestly, I had in my mind that maintenance tick was 1s.

core/src/proxy/block_engine_stage.rs
jedleggett: Checking once per second now.  Will explore other signal mechanisms in followup pr for dynamic back end config changes.

jedleggett: Checking once per second now.  Will explore other signal mechanisms in followup pr for dynamic back end config changes.

core/src/proxy/auth.rs
```diff
@@ -42,206 +40,161 @@ impl Interceptor for AuthInterceptor {
...
+pub async fn maybe_refresh_auth_tokens(
+    auth_service_client: &mut AuthServiceClient<Channel>,
+    datapoint_tokens_generated: &'static str,
+    datapoint_access_refresh: &'static str,
+    auth_uri: &String,
+    access_token: &Arc<Mutex<Token>>,
+    refresh_token: Token,
+    cluster_info: &Arc<ClusterInfo>,
+    connection_timeout: &Duration,
+    num_full_refreshes: &mut u64,
```
buffalu, line 106: i think can pull these &mut num_full_refreshes and num_refresh_access_token out of this method and do datapoints in the caller?

core/src/proxy/block_engine_stage.rs
```diff
@@ -272,42 +330,78 @@ impl BlockEngineStage {
...
+                    Self::handle_block_engine_packets(resp, packet_tx, verified_packet_tx, block_engine_config.trust_packets, &mut block_engine_stats)?;
                 }
                 maybe_bundles = bundle_stream.message() => {
                     Self::handle_block_engine_maybe_bundles(maybe_bundles, bundle_tx, &mut block_engine_stats)?;
                 }
                 _ = metrics_tick.tick() => {
                     block_engine_stats.report();
                     block_engine_stats = BlockEngineStageStats::default();
+
+                    if cluster_info.id() != keypair.pubkey() {
```
buffalu, line 369: do we generate a datapoint anywhere for this? might be helpful

core/src/proxy/block_engine_stage.rs
```diff
@@ -272,42 +330,78 @@ impl BlockEngineStage {
...
+                        &auth_uri_string,
+                        &access_token,
+                        refresh_token.clone(),  // refresh token is consumed
+                        &cluster_info,
+                        connection_timeout,
+                        &mut num_full_refreshes,
+                        &mut num_refresh_access_token)
+                    .await?;
+
+                    if let Some(new_token) = maybe_new_access { *access_token.lock().unwrap() = new_token;}
```
buffalu, line 385: nit: format

core/src/proxy/block_engine_stage.rs
jedleggett: we record proxy errors here
https://github.com/jito-foundation/jito-solana/blob/d83e03076001080cb1d6dae71b70b7caa78f5d89/core/src/proxy/block_engine_stage.rs#L166-L170

would have to look at the error message to distinguish between types of proxy errors, but usually the errors are all the same type for a particular event.

core/src/proxy/auth.rs
jedleggett: that removes the string paramaters as well +1

core/src/proxy/block_engine_stage.rs
jedleggett: i think this is good for logging these

Screenshot_20230425_154231

core/src/proxy/auth.rs
```diff
@@ -42,206 +40,143 @@ impl Interceptor for AuthInterceptor {
...
+    auth_service_client: &mut AuthServiceClient<Channel>,
+    access_token: &Arc<Mutex<Token>>,
+    refresh_token: Token,
+    cluster_info: &Arc<ClusterInfo>,
+    connection_timeout: &Duration,
+) -> crate::proxy::Result<(
+    Option<Token>, // access token
+    Option<Token>, // refresh token
+)> {
+    const REFRESH_WITHIN_S: u64 = 5;
```
buffalu, line 107: nit: should pass this in and should be related to how often refresh happens

buffalu: for instance, if we change refresh period to 1 minute, there could be a period of time where the refresh token is bad.

the caller should be responsible for saying when the refresh should happen within

jedleggett: makes sense

jedleggett: makes sense

core/src/proxy/block_engine_stage.rs
```diff
@@ -272,42 +339,90 @@ impl BlockEngineStage {
...
+        mut auth_client: AuthServiceClient<Channel>,
+        access_token: Arc<Mutex<Token>>,
+        mut refresh_token: Token,
+        keypair: Arc<Keypair>,
+        cluster_info: &Arc<ClusterInfo>,
+        connection_timeout: &Duration,
     ) -> crate::proxy::Result<()> {
         const METRICS_TICK: Duration = Duration::from_secs(1);
         const MAINTENANCE_TICK: Duration = Duration::from_secs(10 * 60);
+        let refresh_within_s: u64 = std::cmp::max(METRICS_TICK.as_secs() / 2, 1);
```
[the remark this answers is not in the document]
jedleggett: whoops, yes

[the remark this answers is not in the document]
jedleggett: whoops, yes

core/src/proxy/auth.rs
```diff
@@ -1,4 +1,6 @@
+use tonic::Code;
```
buffalu, line 1: cargo fmt?

jedleggett: weird, i have it set to run fmt on save, but had to manually run to get it to fix this.

buffalu (approved):

PR #294, opened 2023-04-11
Replay using a tx dependency graph

buffalu:
#### Problem
- The current replay algorithm is less-than-optimal. It inefficiently batches transactions and shows performance issues compared to a more novel algorithm.

#### Summary of Changes
- When it comes time to replay a chunk of transactions, a dependency graph is based that represents "what transactions need to execute before this transaction can execute".
- Any transactions not blocked (not blocked = all dependent transactions have executed) are scheduled.
- When the response is returned, that transaction is marked as processed and the loop continues.

Note: I would strongly encourage against running this algorithm on a voting validator for reliability and performance reasons. It hasn't been stress tested and the effects on the rest of the network haven't been examined. It should be suitable for RPCs where performance is a concern.

**Comparison to another private RPC running v1.13-fast-replay upstream branch**
Screenshot 2023-04-11 at 5 08 48 PM


core/src/replay_stage.rs
```diff
@@ -1659,12 +1671,16 @@ impl ReplayStage {
...
+        handle: &BankTransactionExecutorHandle,
     ) -> result::Result<usize, BlockstoreProcessorError> {
         let tx_count_before = bank_progress.replay_progress.num_txs;
         // All errors must lead to marking the slot as dead, otherwise,
         // the `check_slot_agrees_with_cluster()` called by `replay_active_banks()`
         // will break!
-        blockstore_processor::confirm_slot(
+
+        // loop over until no more entries to process
+        // TODO (LB): probably wanna add a time limit here
```
segfaultdoc, line 1682: can expand on why we should add a time limit?

core/src/replay_stage.rs
```diff
@@ -1659,12 +1671,16 @@ impl ReplayStage {
...
+        while blockstore_processor::confirm_slot(
```
segfaultdoc, line 1683: Just to clarify, this will now attempt to replay all available entries for a bank vs. jumping back to call-site after replaying one set of entries?

ledger/src/bank_transaction_executor.rs
```diff
@@ -0,0 +1,547 @@
+use {
```
segfaultdoc, line 1: this code mostly unchanged?

ledger/src/blockstore_processor.rs
```diff
@@ -111,268 +78,198 @@ fn first_err(results: &[Result<()>]) -> Result<()> {
...
-            execute_cost_units.saturating_add(timing.accumulated_units / timing.count as u64);
-        trace!("aggregated execution cost of {:?} {:?}", program_id, timing);
-    }
-    execute_cost_units
+    tx_account_locks
+        .iter()
+        .enumerate()
+        .map(|(idx, account_locks)| {
+            let mut dep_graph: IntSet<usize> = IntSet::default();
+            let readlock_accs = account_locks.writable.iter();
```
segfaultdoc, line 113: var naming kind of confusing here how about something like:
```
let write_locks = account_locks.writable.iter();
let read_and_write_locks = account_locks
                .readonly
                .iter()
                .chain(account_locks.writable.iter());
```
The for-loops below translate to something like:
"Find all transactions that read-lock my write-locks, filter for transactions that come before me."
"Find all transactions that write-lock my write OR read locks, filter for transactions that come before me."

ledger/src/blockstore_processor.rs
```diff
@@ -111,268 +78,198 @@ fn first_err(results: &[Result<()>]) -> Result<()> {
...
-            collect_token_balances(bank, batch, &mut mint_decimals, None)
-        } else {
-            vec![]
-        };
+    let receiver = tx_executor_handle.response_receiver();
 
-        let token_balances =
-            TransactionTokenBalancesSet::new(pre_token_balances, post_token_balances);
+    let mut processing_states: Vec<State> = vec![State::Blocked; dependency_graph.len()];
+    let mut signature_indices: HashMap<&Signature, usize> =
```
segfaultdoc, line 174:
```suggestion
    let signature_indices: HashMap<&Signature, usize> = pending_transactions
        .iter()
        .enumerate()
        .map(|(idx, tx)| (tx.signature(), idx))
        .collect();
```

ledger/src/blockstore_processor.rs
```diff
@@ -111,268 +78,198 @@ fn first_err(results: &[Result<()>]) -> Result<()> {
...
+    while !is_done {
+        is_done = true;
+        for idx in 0..processing_states.len() {
+            match processing_states[idx] {
+                State::Blocked => {
+                    is_done = false;
+
+                    // if all the dependent txs are executed, this transaction can be scheduled for
+                    // execution.
+                    if dependency_graph[idx]
```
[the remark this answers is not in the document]
segfaultdoc: nvm `all` returns true on empty iter

segfaultdoc (approved):
small suggestions, lgtm

core/src/replay_stage.rs
```diff
@@ -893,6 +902,9 @@ impl ReplayStage {
                         retransmit_not_propagated_time.as_us(),
                     );
                 }
+
+                drop(handle);
```
esemeniuc, line 906: why drop handle?

ledger/src/bank_transaction_executor.rs
```diff
@@ -0,0 +1,547 @@
...
+    };
+
+    let pre_process_units: u64 = aggregate_total_execution_units(timings);
+
+    let (tx_results, balances) = batch.bank().load_execute_and_commit_transactions(
+        batch,
+        MAX_PROCESSING_AGE,
+        transaction_status_sender.is_some(),
+        transaction_status_sender.is_some(),
+        transaction_status_sender.is_some(),
```
esemeniuc, line 308:
```suggestion
        record_token_balances,
        record_token_balances,
        record_token_balances,
```

ledger/src/bank_transaction_executor.rs
```diff
@@ -0,0 +1,547 @@
...
+
+            if response_sender
+                .send(BankTransactionExecutionResponse {
+                    result: execution_result,
+                    timings,
+                    signature,
+                })
+                .is_err()
+            {
+                warn!("error sending back results");
```
esemeniuc, line 277:
```suggestion
                warn!("error sending back results for signature: {signature}");
```

ledger/src/bank_transaction_executor.rs
```diff
@@ -0,0 +1,547 @@
...
+
+        debug!(
+            "bank {} executed a batch, number of transactions {}, total execute cu {}, remaining block cost cap {}",
+            bank.slot(),
+            batch.sanitized_transactions().len(),
+            execution_cost_units,
+            remaining_block_cost_cap,
+        );
+
+        if remaining_block_cost_cap == 0_u64 {
```
esemeniuc, line 330:
```suggestion
        if remaining_block_cost_cap == 0 {
```

ledger/src/bank_transaction_executor.rs
```diff
@@ -0,0 +1,547 @@
...
+            warn!(
+                "Unexpected validator error: {:?}, transaction: {:?}",
+                err, transaction
+            );
+            datapoint_error!(
+                "validator_process_entry_error",
+                (
+                    "error",
+                    format!("error: {:?}, transaction: {:?}", err, transaction),
+                    String
```
esemeniuc, line 469:
```suggestion
            datapoint_error!(
                "validator_process_entry_error",
                ("error", err.to_string(), String),
                ("transaction", format!("{:?}", transaction), String)
            );
```

core/src/replay_stage.rs
buffalu: yeah

ledger/src/bank_transaction_executor.rs
buffalu: some of it is copied from blockstore_processor to avoid circulaar dep, some of it new. most fns at the bottom are copied

ledger/src/bank_transaction_executor.rs
buffalu: copied from blockstore_processor.rs, going to leave as is

ledger/src/bank_transaction_executor.rs
buffalu: copied from blockstore_processor.rs, going to leave as is

ledger/src/bank_transaction_executor.rs
buffalu: copied from blockstore_processor.rs, going to leave as is

ledger/src/bank_transaction_executor.rs
```diff
@@ -0,0 +1,547 @@
...
+                )
+            );
+        }
```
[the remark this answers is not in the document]
buffalu: copied from blockstore_processor.rs, going to leave as is

esemeniuc (approved):

core/src/replay_stage.rs
```diff
@@ -437,6 +441,10 @@ impl ReplayStage {
                 };
                 let in_vote_only_mode = bank_forks.read().unwrap().get_vote_only_mode_signal();
 
+                // TODO (LB): dont do * 2
```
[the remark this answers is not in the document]
buffalu:
```
lazy_static! {
    // reduce the number of threads each pool is allowed to half the cpu core count, to avoid rayon
    // hogging cpu
    static ref MAX_RAYON_THREADS: usize =
            env::var("SOLANA_RAYON_THREADS").ok()
            .and_then(|num_threads| num_threads.parse().ok())
            .unwrap_or_else(|| num_cpus::get() as usize / 2)
            .max(1);
}

pub fn get_thread_count() -> usize {
    *MAX_RAYON_THREADS
}
```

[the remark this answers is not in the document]
buffalu: gonna use num_cpus crate directly instead

core/src/replay_stage.rs
buffalu: senders are cloned when you create a handle. the bank_tx_execution.join() will drop bank_tx_execution.sender and once all the senders are dropped, bank_tx_execution will shutdown

core/src/replay_stage.rs
buffalu: ill add now

ledger/src/blockstore_processor.rs
```diff
@@ -111,268 +78,198 @@ fn first_err(results: &[Result<()>]) -> Result<()> {
...
-    batch: &TransactionBatch,
-    fee_collection_results: Vec<Result<()>>,
-) -> Option<(Result<()>, Signature)> {
-    let mut first_err = None;
-    for (result, transaction) in fee_collection_results
+// for each index, builds a transaction dependency graph of indices that need to execute before
+// the current one.
+// The returned Vec<HashSet<usize>> is a 1:1 mapping for the indices that need to be executed
+// before that index can be executed
+fn build_dependency_graph(tx_account_locks: &[TransactionAccountLocks]) -> Vec<IntSet<usize>> {
```
ryoqun, line 85: @buffalu just hi. :)

fyi, i just copied this dep graph logics to https://github.com/solana-labs/solana/pull/31239#discussion_r1205565836 for bench purposes (not on production code-path, for now)

that massive pr contains bigger-yet-optimized `BankTransactionExecutor` (i.e. thread pool for executing unbatched txes), imho. so, you might be interested.

that pr's general mechanism will be plumbed with banking stage. so, it's too early for you to comment on it, but i hope any jito-specific block ordering with bare Txes and Bundles will be more straight forward to implement (including git rebases)

PR #295, opened 2023-04-16
account for case where TDA exists but not allocated

segfaultdoc:
TDAs may exist in the bank but not have been initialized yet thru the program invocation. Account for this rather than panic.

### DIFF
--- a/tip-distributor/src/stake_meta_generator_workflow.rs
+++ b/tip-distributor/src/stake_meta_generator_workflow.rs
@@ -208,12 +208,12 @@ pub fn generate_stake_meta_collection(
                 bank.epoch(),
             )
             .0;
-            let tda = bank
-                .get_account(&tip_distribution_pubkey)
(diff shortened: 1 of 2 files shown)


tip-distributor/src/stake_meta_generator_workflow.rs
```diff
@@ -208,12 +208,12 @@ pub fn generate_stake_meta_collection(
                 bank.epoch(),
             )
             .0;
-            let tda = bank
-                .get_account(&tip_distribution_pubkey)
-                .map(|mut account_data| {
-                    let tip_distribution_account =
-                        TipDistributionAccount::try_deserialize(&mut account_data.data())
-                            .expect("deserialized TipDistributionAccount");
+            let tda = if let Some(mut account_data) = bank.get_account(&tip_distribution_pubkey) {
```
buffalu, line 211: nit: some of this stuff can use `?` operator, but nbd

PR #298, opened 2023-04-26
limit number of concurrent rpc calls

segfaultdoc:
can overload system when spawning too many concurrent network calls

### DIFF
--- a/tip-distributor/src/lib.rs
+++ b/tip-distributor/src/lib.rs
@@ -26,7 +26,7 @@ use {
         pubkey::Pubkey,
         signature::{Keypair, Signature},
         stake_history::Epoch,
-        transaction::Transaction,
+        transaction::{Transaction, TransactionError::AlreadyProcessed},
     },
     std::{
         collections::HashMap,
@@ -522,7 +522,11 @@ pub async fn sign_and_send_transactions_with_retries(
                     }
                     Err(e) => {
                         error!("error sending transaction {sig:?} error: {e:?}");
-                        Err(e)
+                        if e == AlreadyProcessed {
+                            Ok(())
+                        } else {
+                            Err(e)
+                        }
                     }
                 };
 
--- a/tip-distributor/src/lib.rs
+++ b/tip-distributor/src/lib.rs
@@ -522,7 +522,7 @@ pub async fn sign_and_send_transactions_with_retries(
                     }
                     Err(e) => {
                         error!("error sending transaction {sig:?} error: {e:?}");
-                        if e == AlreadyProcessed {
+                        if let ErrorKind::TransactionError(AlreadyProcessed) = e.kind {
                             Ok(())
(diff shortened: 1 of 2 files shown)


tip-distributor/src/lib.rs
```diff
@@ -513,12 +521,17 @@ pub async fn sign_and_send_transactions_with_retries(
                     }
                     Err(e) => {
                         error!("error sending transaction {sig:?} error: {e:?}");
-                        Err(e)
+                        if let ErrorKind::TransactionError(AlreadyProcessed) = e.kind {
+                            Ok(())
+                        } else {
+                            Err(e)
```
esemeniuc, line 527: maybe move line 523 to be inside the block so we aren't spammed with already processed msgs

esemeniuc (approved):

tip-distributor/src/lib.rs
```diff
@@ -494,14 +499,17 @@ pub async fn sign_and_send_transactions_with_retries(
...
-            .map(|(sig, tx)| async move {
+        let futs = signatures_to_transactions.iter().map(|(sig, tx)| {
+            let semaphore = semaphore.clone();
+            async move {
+                let permit = semaphore.clone().acquire_owned().await.unwrap();
                 let res = match rpc_client.send_transaction(tx).await {
                     Ok(_sig) => {
                         info!("sent transaction: {_sig:?}");
+                        drop(permit);
                         sleep(Duration::from_secs(5)).await;
```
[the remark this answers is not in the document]
segfaultdoc: the sleep is so we give the tx enough time to confirm

tip-distributor/src/lib.rs
```diff
@@ -494,14 +499,17 @@ pub async fn sign_and_send_transactions_with_retries(
                 });
         }
 
-        let futs = signatures_to_transactions
-            .iter()
-            .map(|(sig, tx)| async move {
+        let futs = signatures_to_transactions.iter().map(|(sig, tx)| {
```
buffalu, line 502: wb single threading sends here (sends are over localhost so super fast) then single threading confirmations (same wrt speed)

tip-distributor/src/lib.rs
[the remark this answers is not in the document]
segfaultdoc: the sleep is so we give the tx enough time to confirm

tip-distributor/src/lib.rs
buffalu: alternatively, could limit number of tokio worker threads

buffalu: and not use semaphore

esemeniuc: could do:
```
let mut rt = runtime::Builder::new()
    .core_threads(4)
    .build()
    .unwrap();

```

PR #306, opened 2023-05-16
Admin Shred Receiver Change

jedleggett:
#### Problem
Currently, an additional shred receiver can be specified at startup through a cli argument.  However, this cannot be changed at runtime.

#### Summary of Changes
Add an admin rpc call to change the shred receiver address at runtime, and update the broadcast and retransmit stages to accept a shared address.


core/src/broadcast_stage.rs
```diff
@@ -86,7 +86,7 @@ impl BroadcastStageType {
         blockstore: Arc<Blockstore>,
         bank_forks: Arc<RwLock<BankForks>>,
         shred_version: u16,
-        shred_receiver_addr: Option<SocketAddr>,
+        shred_receiver_addr: Arc<RwLock<Option<SocketAddr>>>,
```
buffalu, line 89: can we use Option<Arc<RwLock<SocketAddr>> here? doing Arc::new(RwLock::new(None)) is dirty

validator/src/admin_rpc_service.rs
```diff
@@ -206,6 +208,9 @@ pub trait AdminRpc {
         expected_heartbeat_interval_ms: u64,
         max_failed_heartbeats: u64,
     ) -> Result<()>;
+
+    #[rpc(meta, name = "setBlockEngineConfig")]
```
buffalu, line 212: setShredReceiverAddress

core/src/broadcast_stage.rs
buffalu: ah nvmd, you want to be able to turn it on + off dynamically. will leave comment here in case someone else gets confused :)

buffalu: should we just use a mutex tho?

validator/src/admin_rpc_service.rs
```diff
@@ -382,6 +387,21 @@ impl AdminRpc for AdminRpcImpl {
         }
     }
 
+    fn set_shred_receiver_address(&self, meta: Self::Metadata, addr: String) -> Result<()> {
+        let shred_receiver_address = if addr.is_empty() {
+            None
+        } else {
+            Some(SocketAddr::from_str(&addr).map_err(|_| {
+                jsonrpc_core::error::Error::invalid_params("invalid shred receiver address")
```
buffalu, line 395: add bad addr provided string

validator/src/main.rs
```diff
@@ -2057,6 +2057,18 @@ pub fn main() {
...
         )
+        .subcommand(
+            SubCommand::with_name("set-shred-receiver-address")
+                .about("Changes shred receiver address")
+                .arg(
+                    Arg::with_name("shred_receiver_address")
+                        .long("shred-receiver-address")
+                        .value_name("SHRED_RECEIVER_ADDRESS")
+                        .takes_value(true)
+                        .help("Shred receiver listening address. Set to empty string to disable.")
```
buffalu, line 2068: this is kinda misleading/shred receiver address kinda misleading in general. wish we would've named it shred forwarder address... can you make this help string more clear its an address we forward shreds to?

core/src/broadcast_stage.rs
jedleggett: looks like it's written to infrequently, and read frequently in 2 different threads (broadcast_stage, and retransmit_stage) 

so, i think it should stay, but not 100% sure how often the reads are hit

buffalu: ah yeah great point. lets keep as is

buffalu (approved):

buffalu (approved):

segfaultdoc (approved):

PR #304, opened 2023-05-10
Set backend config from admin rpc

jedleggett:
#### Problem
Configuration of Block Engine and Relayer connections is currently only possible through command line parameters.  However, if these parameters are not included at startup or are entered incorrectly, a validator restart is required to change them.  

#### Summary of Changes
Add admin rpc functions to set block engine and relayer configurations and update auth to detect changes and re-connect.

**Note:** This PR removes the following deprecated cli and admin rpc args:
- block_engine_address, 
- block_engine_auth_service_address,
- relayer_address,
- relayer_auth_service_address

They were superseded by block_engine_url and relayer_url.

PR #308, opened 2023-05-20
Jl/backport 306 v1.14

jedleggett:
#### Summary of Changes
Backport https://github.com/jito-foundation/jito-solana/pull/306

### DIFF
--- a/validator/src/main.rs
+++ b/validator/src/main.rs
@@ -1946,7 +1946,7 @@ pub fn main() {
                 .long("shred-receiver-address")
                 .value_name("SHRED_RECEIVER_ADDRESS")
                 .takes_value(true)
-                .help("Shred receiver listening address")
+                .help("Validator will forward all shreds to this address in addition to normal turbine operation. Set to empty string to disable.")
         )
         .arg(
             Arg::with_name("log_messages_bytes_limit")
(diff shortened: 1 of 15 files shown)


buffalu (approved):

validator/src/main.rs
```diff
@@ -1946,7 +1946,7 @@ pub fn main() {
                 .long("shred-receiver-address")
                 .value_name("SHRED_RECEIVER_ADDRESS")
                 .takes_value(true)
-                .help("Shred receiver listening address")
+                .help("Validator will forward all shreds to this address in addition to normal turbine operation. Set to empty string to disable.")
```
segfaultdoc, line 1949: nit: omit to disable?

validator/src/main.rs
```diff
@@ -2169,6 +2169,18 @@ pub fn main() {
...
+        .subcommand(
+            SubCommand::with_name("set-shred-receiver-address")
+                .about("Changes shred receiver address")
+                .arg(
+                    Arg::with_name("shred_receiver_address")
+                        .long("shred-receiver-address")
+                        .value_name("SHRED_RECEIVER_ADDRESS")
+                        .takes_value(true)
+                        .help("Validator will forward all shreds to this address in addition to normal turbine operation. Set to empty string to disable.")
+                        .required(true)
```
segfaultdoc, line 2181: must it be required?

segfaultdoc (approved):

validator/src/main.rs
jedleggett: i think in the subcommand, we want it required,

see discussion here
https://github.com/jito-foundation/jito-solana/pull/304#discussion_r1197895880

validator/src/main.rs
jedleggett: yeah, if they omit at startup it will be disabled, will fix help

PR #334, opened 2023-08-14
adds back max-rpc-payload-size option

segfaultdoc:
This got removed in feature branches but is still in master. Need it for larger bundles.


buffalu (approved):

buffalu:
lgtm after CI passes

PR #392, opened 2023-10-04
Batched Tip Dist

esemeniuc:
Notes:
- Fetches accounts in batches of 100
- Retry fetching accounts
- Avoid panic when, forcing account refetch
- Avoid memory bloat with new blockhashes creating duplicate signatures
- Log datapoints
- Balance check delayed (avoid 2x balance requirement)
- Increase CU limit for batching instructions


jito-programs
segfaultdoc, line 1: make sure to rebase your commit on `submodule` and use that

tip-distributor/src/claim_mev_workflow.rs
```diff
@@ -32,121 +39,343 @@ pub enum ClaimMevError {
...
+            }
+        };
+
+        let account = match TipDistributionAccount::try_deserialize(&mut account.data.as_slice()) {
+            Ok(a) => a,
+            Err(e) => {
+                datapoint_warn!(
+                    "claim_mev_workflow-account_error",
+                    ("pubkey", pubkey.to_string(), String),
+                    ("pubkey_type", "tip_distribution_account", String),
```
segfaultdoc, line 113:
```suggestion
                    ("account_type", "tip_distribution_account", String),
```

tip-distributor/src/claim_mev_workflow.rs
```diff
@@ -32,121 +39,343 @@ pub enum ClaimMevError {
...
+                    ("error", 1, i64),
+                    ("err_type", "deserialize_tip_distribution_account", String),
```
segfaultdoc, line 115: seems redundant, how about just changing "error" to be "deser_error_count"?

tip-distributor/src/claim_mev_workflow.rs
```diff
@@ -32,121 +39,343 @@ pub enum ClaimMevError {
...
+    .unwrap()
+    .into_iter()
+    .filter_map(|(pubkey, maybe_account)| {
+        let account = match maybe_account {
+            Some(a) => a,
+            None => {
+                datapoint_warn!(
+                    "claim_mev_workflow-account_error",
+                    ("pubkey", pubkey.to_string(), String),
+                    ("pubkey_type", "claimant", String),
```
segfaultdoc, line 142:
```suggestion
                    ("account_type", "claimant", String),
```

tip-distributor/src/claim_mev_workflow.rs
esemeniuc: the rationale for "error", 1 is mainly for the charts, and "err_type" for debugging

jito-programs
esemeniuc: thank you

tip-distributor/src/claim_mev_workflow.rs
```diff
@@ -32,121 +39,343 @@ pub enum ClaimMevError {
...
+            // make sure not previously claimed
+            match claim_statuses.get(&node.claim_status_pubkey) {
+                Some(Some(_account)) => {
+                    debug!(
+                        "Claim status account already exists (already paid out). Skipping pubkey: {:?}.", node.claim_status_pubkey,
+                    );
+                    already_claimed_count = already_claimed_count.checked_add(1).unwrap();
                     continue;
                 }
+                None => panic!(
```
segfaultdoc, line 198: can you document why this would happen? preferably would like to avoid panicking

segfaultdoc: seems like would happen if rpc error, in that case we should fail before we get to this step

esemeniuc: yes, this unrap() should prevent this case: https://github.com/jito-foundation/jito-solana/pull/392/files#diff-f2fd8703d48635e7d3a0ec5182e43a1dde5ae307fa6f637d2025162d692c24b3R162

tip-distributor/src/claim_mev_workflow.rs
```diff
@@ -32,121 +39,343 @@ pub enum ClaimMevError {
...
+            failed_transactions.len()
+        );
+    }
 
     Ok(())
 }
+async fn get_batched_accounts(
+    rpc_client: &RpcClient,
+    pubkeys: Vec<Pubkey>,
+) -> solana_rpc_client_api::client_error::Result<HashMap<Pubkey, Option<Account>>> {
```
segfaultdoc, line 361: why not just `HashMap<Pubkey, Account>` and if DNE don't include the entry

esemeniuc: rn looking to see it panic if i have a logic error. will switch over after testing :)

PR #403, opened 2023-10-10
fix sim bundle

segfaultdoc:


accounts-db/src/ancient_append_vecs.rs
```diff
@@ -1462,10 +1462,8 @@ pub mod tests {
                             .map(|store| db.get_unique_accounts_from_storage(store))
                             .collect::<Vec<_>>();
 
-                        let accounts_per_storage = infos
-                            .iter()
-                            .zip(original_results.into_iter())
-                            .collect::<Vec<_>>();
+                        let accounts_per_storage =
```
buffalu, line 1465: we should remove any changes to core solana codebase for lints/clippy stuff. makes the diff super hard to parse

buffalu: also increases likelihood of merge conflict in the future

PR #487, opened 2023-11-29
 [JIT-1712] Fix BundleStorage rebuffering

esemeniuc:
#### Problem
Bundles may be rebuffered for longer than necessary when cost model checks fail.

#### Summary of Changes
- Add TTL to bundles to limit amount of retries

Fixes #

### DIFF
--- a/core/src/banking_stage/unprocessed_transaction_storage.rs
+++ b/core/src/banking_stage/unprocessed_transaction_storage.rs
@@ -263,10 +263,35 @@ impl UnprocessedTransactionStorage {
     }
 
     pub fn new_bundle_storage(
-        unprocessed_bundle_storage: VecDeque<ImmutableDeserializedBundle>,
-        cost_model_failed_bundles: VecDeque<ImmutableDeserializedBundle>,
+        unprocessed_bundle_storage: VecDeque<(
+            u8, /* attempts remaining */
+            ImmutableDeserializedBundle,
+        )>,
+        cost_model_failed_bundles: VecDeque<(
+            u8, /* attempts remaining */
+            ImmutableDeserializedBundle,
+        )>,
+    ) -> Self {
+        Self::new_bundle_storage_ttl(
+            unprocessed_bundle_storage,
+            cost_model_failed_bundles,
+            BundleStorage::MAX_BUNDLE_RETRY_ATTEMPTS,
+        )
+    }
+
+    pub fn new_bundle_storage_ttl(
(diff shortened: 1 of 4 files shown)


core/src/bundle_stage.rs
```diff
@@ -301,80 +301,103 @@ impl BundleStage {
...
+        if let Err(RecvTimeoutError::Disconnected) = bundle_receiver.receive_and_buffer_bundles(
+            unprocessed_bundle_storage,
+            bundle_stage_metrics,
+            bundle_stage_leader_metrics,
+        ) {
+            return Err(RecvTimeoutError::Disconnected);
         }
+
+        let bundle_storage = unprocessed_bundle_storage.bundle_storage().unwrap();
+        bundle_stage_metrics.increment_current_buffered_bundles_count(
```
segfaultdoc, line 384: inherited a bug with these metrics? seems like should be setters rather than incrementing. cc: @buffalu

esemeniuc: updated

core/src/banking_stage/unprocessed_transaction_storage.rs
```diff
@@ -263,10 +263,35 @@ impl UnprocessedTransactionStorage {
...
+        unprocessed_bundle_storage: VecDeque<(
+            u8, /* attempts remaining */
+            ImmutableDeserializedBundle,
+        )>,
+        cost_model_failed_bundles: VecDeque<(
+            u8, /* attempts remaining */
+            ImmutableDeserializedBundle,
+        )>,
+    ) -> Self {
+        Self::new_bundle_storage_ttl(
```
jedleggett, line 275: nit: why does this need to be a separate function call?

esemeniuc: this was to avoid modifying all the call sites with the TTL info

core/src/banking_stage/unprocessed_transaction_storage.rs
```diff
@@ -263,10 +263,35 @@ impl UnprocessedTransactionStorage {
...
+            unprocessed_bundle_storage,
+            cost_model_failed_bundles,
+            BundleStorage::MAX_BUNDLE_RETRY_ATTEMPTS,
+        )
+    }
+
+    pub fn new_bundle_storage_ttl(
```
segfaultdoc, line 282: what's the ttl?

PR #489, opened 2023-11-30
[JIT-1713] Fix bundle's blockspace preallocation

esemeniuc:
#### Problem

Reserved bundle space is not correctly preallocating space. `BundleReservedSpaceManager::is_in_reserved_tick_period` checks `bank.tick_height()` which represents the number of ticks since genesis, not since the start of the slot. `is_in_reserved_tick_period` will only work properly in the first slot after genesis.

#### Summary of Changes
- Use modular division to correctly get the ticks since start of slot

### DIFF
--- a/core/src/bundle_stage/bundle_reserved_space_manager.rs
+++ b/core/src/bundle_stage/bundle_reserved_space_manager.rs
@@ -192,7 +192,7 @@ mod tests {
     #[test]
     fn test_block_limits_after_first_slot() {
         const BUNDLE_BLOCK_COST_LIMITS_RESERVATION: u64 = 100;
-
+        const RESERVED_TICKS: u64 = 5;
         let genesis_config_info = create_genesis_config(100);
         let bank = Arc::new(Bank::new_for_tests(&genesis_config_info.genesis_config));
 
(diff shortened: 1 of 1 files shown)


core/src/bundle_stage/bundle_reserved_space_manager.rs
```diff
@@ -186,4 +188,41 @@ mod tests {
...
+
+        for _ in 0..genesis_config_info.genesis_config.ticks_per_slot {
+            bank.register_tick(&Hash::default());
+        }
+        assert!(bank.is_complete());
+        bank.freeze();
+
+        let bank1 = Arc::new(Bank::new_from_parent(bank, &Pubkey::default(), 1));
+
+        // Prints: bank1: 1, tick_height: 64, max_tick_height: 128
```
segfaultdoc, line 207: del

core/src/bundle_stage/bundle_reserved_space_manager.rs
```diff
@@ -186,4 +188,41 @@ mod tests {
...
+        //     bank1.max_tick_height()
+        // );
+
+        let block_cost_limits = bank1.read_cost_tracker().unwrap().block_cost_limit();
+        let mut reserved_space = BundleReservedSpaceManager::new(
+            block_cost_limits,
+            BUNDLE_BLOCK_COST_LIMITS_RESERVATION,
+            5,
+        );
+        reserved_space.tick(&bank1);
```
segfaultdoc, line 221: can you add a test case that tests the cost tracker reverting back to original block cost limits, after it's advanced enough ticks?

buffalu (approved):

PR #492, opened 2023-12-01
[JIT-1706] Propagate heartbeat configs to relayer_config when updated thru CLI

esemeniuc:
#### Problem
The code to change the heartbeat interval and max failed heart beats through relayer_config on CLI is  not propagated to FetchStageManager.

#### Summary of Changes
- Wire up relayer_config
- Restart loop when config updated
- Add better error handling

Fixes #


core/src/proxy/fetch_stage_manager.rs
```diff
@@ -37,13 +39,23 @@ impl FetchStageManager {
...
     ) -> Self {
-        let t_hdl = Self::start(
-            cluster_info,
-            heartbeat_rx,
-            packet_intercept_rx,
-            packet_tx,
-            exit,
-        );
+        let t_hdl = Builder::new()
+            .name("fetchStageManager".to_string())
```
segfaultdoc, line 43:
```suggestion
            .name("solFetchStageManager".to_string())
```

core/src/proxy/fetch_stage_manager.rs
```diff
@@ -61,97 +73,104 @@ impl FetchStageManager {
...
-        Builder::new().name("fetch-stage-manager".into()).spawn(move || {
-            let my_fallback_contact_info = cluster_info.my_contact_info();
+        global_relayer_config: &Arc<Mutex<RelayerConfig>>,
+        cluster_info: &Arc<ClusterInfo>,
+        heartbeat_rx: &Receiver<HeartbeatEvent>,
+        packet_intercept_rx: &Receiver<PacketBatch>,
+        packet_tx: &Sender<PacketBatch>,
+        exit: &Arc<AtomicBool>,
+    ) -> ProxyResult<()> {
+        let my_fallback_contact_info = cluster_info.my_contact_info();
```
segfaultdoc, line 83: can you add a comment on what this is and why we don't care if `cluster_info` gets changed via the admin rpc?

core/src/proxy/fetch_stage_manager.rs
```diff
@@ -61,97 +73,104 @@ impl FetchStageManager {
...
-                        datapoint_info!(
-                            "relayer-heartbeat",
-                            ("fetch_stage_packets_forwarded", packets_forwarded, i64),
-                            ("heartbeats_received", heartbeats_received, i64),
-                        );
-
+                }
+                recv(relayer_config_tick) -> _ => {
+                    if local_relayer_config != global_relayer_config.lock().unwrap().clone() {
+                        return Err(ProxyError::RelayerConfigChanged);
```
segfaultdoc, line 162: doesn't seem like this should be an error? wdyt we just update the relevant vars if this changes

esemeniuc: we use Err to reconnect

esemeniuc: we use the Err to signal the loop to reconnect. the return type is `()`. wdyt makes more sense?

core/src/proxy/fetch_stage_manager.rs
esemeniuc: gonna truncate since 15 char is limit

core/src/proxy/fetch_stage_manager.rs
esemeniuc: added a comment on the contact info. not sure what you would like to include wrt not caring if cluster_info gets changed

segfaultdoc: So let's say validator starts up advertising relayer as it's TPU. Now let's say the following happens:
1. Relayer crashes and misses heartbeat
2. Validator falls back to original tpu
3. Validator changes tpu socket via admin rpc
4. Relayer reconnects
5. Fails heartbeat
6. Validator falls back to the original socket even though they had changed it in step 3

PR #495, opened 2023-12-04
[JIT-1812] Fix blocking mutexs

esemeniuc:
#### Problem
Sync mutexs can hold up async runtime. Note likely to happen in production, this is preventative.

#### Summary of Changes
- Spawn mutex lock on blocking-specific threads

Todo: update after #492 lands

### DIFF
--- a/core/src/proxy/block_engine_stage.rs
+++ b/core/src/proxy/block_engine_stage.rs
@@ -35,7 +35,10 @@ use {
         thread::{self, Builder, JoinHandle},
         time::Duration,
     },
-    tokio::time::{interval, sleep, timeout},
+    tokio::{
+        task,
+        time::{interval, sleep, timeout},
+    },
     tonic::{
         codegen::InterceptedService,
         transport::{Channel, Endpoint},
@@ -148,7 +151,12 @@ impl BlockEngineStage {
         while !exit.load(Ordering::Relaxed) {
             // Wait until a valid config is supplied (either initially or by admin rpc)
             // Use if!/else here to avoid extra CONNECTION_BACKOFF wait on successful termination
-            let local_block_engine_config = block_engine_config.lock().unwrap().clone();
+            let local_block_engine_config = {
+                let block_engine_config = block_engine_config.clone();
+                task::spawn_blocking(move || block_engine_config.lock().unwrap().clone())
+                    .await
+                    .unwrap()
+            };
             if !Self::is_valid_block_engine_config(&local_block_engine_config) {
                 sleep(CONNECTION_BACKOFF).await;
             } else if let Err(e) = Self::connect_auth_and_stream(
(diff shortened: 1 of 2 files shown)


core/src/proxy/block_engine_stage.rs
```diff
@@ -148,7 +151,13 @@ impl BlockEngineStage {
         while !exit.load(Ordering::Relaxed) {
             // Wait until a valid config is supplied (either initially or by admin rpc)
             // Use if!/else here to avoid extra CONNECTION_BACKOFF wait on successful termination
-            let local_block_engine_config = block_engine_config.lock().unwrap().clone();
+            let local_block_engine_config = {
+                let block_engine_config = block_engine_config.clone();
+                task::spawn_blocking(move || block_engine_config.lock().unwrap().clone())
+                    .await
+                    .map_err(ProxyError::TokioJoinError)
```
segfaultdoc, line 158: i dnt think we need this error. if this panicks then some other thread had panicked and poisoned the lock, seems pretty idiomatic to just unwrap here

segfaultdoc: remove map_err and just unwrap

core/src/proxy/block_engine_stage.rs
```diff
@@ -400,7 +415,10 @@ impl BlockEngineStage {
                         return Err(ProxyError::AuthenticationConnectionError("validator identity changed".to_string()));
                     }
 
-                    if *global_config.lock().unwrap() != *local_config {
+                    let global_config = global_config.clone();
+                    if *local_config != task::spawn_blocking(move || global_config.lock().unwrap().clone())
+                        .await
+                        .map_err(ProxyError::TokioJoinError)? {
```
segfaultdoc, line 421: rm map_err and unwrap?

core/src/proxy/block_engine_stage.rs
```diff
@@ -419,7 +437,11 @@ impl BlockEngineStage {
                             ("url", &local_config.block_engine_url, String),
                             ("count", num_refresh_access_token, i64),
                         );
-                        *access_token.lock().unwrap() = new_token;
+
+                        let access_token = access_token.clone();
+                        task::spawn_blocking(move || *access_token.lock().unwrap() = new_token)
+                            .await
+                            .map_err(ProxyError::TokioJoinError)?;
```
segfaultdoc, line 444: ditto

core/src/proxy/block_engine_stage.rs
buffalu: there's a panic handler in the client right?

segfaultdoc: agree on generally avoiding unwraps, but in this case a panic would mean some other thread panicked and poisoned this lock. this seems fine to me and idiomatic across rust codebases

esemeniuc: Also the JoinError only occurs when the tokio task is cancelled, or the sync code panics: https://docs.rs/tokio/latest/tokio/task/struct.JoinError.html

jedleggett (approved):

PR #561, opened 2024-01-14
fix RPC bug

segfaultdoc:


core/src/bundle_stage/bundle_packet_receiver.rs
```diff
@@ -196,21 +197,33 @@ mod tests {
...
-                        ))
+                        ));
+                        (
+                            SanitizedTransaction::try_create(
+                                transaction.clone(),
+                                MessageHash::Compute,
+                                None,
+                                SimpleAddressLoader::Disabled,
+                            )
+                            .unwrap(),
```
esemeniuc, line 216: do we want to filter_map and avoid unwrapping?

core/src/bundle_stage/bundle_consumer.rs
```diff
@@ -983,21 +986,34 @@ mod tests {
...
-                        ))
+                        ));
+                        (
+                            SanitizedTransaction::try_create(
+                                transaction.clone(),
+                                MessageHash::Compute,
+                                None,
+                                SimpleAddressLoader::Disabled,
+                            )
+                            .unwrap(),
```
esemeniuc, line 1005: same, filter_map and avoid unwrap?

PR #592, opened 2024-02-07
Return parsed token accounts in `simulateBundle`

mcintyre94:
#### Problem

If you call `simulateBundle` and pass a token account in `pre/postExecutionAccountsConfigs`, then the token accounts are never returned in JSON. You will always get the base64 encoded data.

This is caused by attempting to encode the data without fetching the number of decimals for the mint, which causes a fallback to base64 data.

This was fixed for `simulateTransaction` upstream, but `simulateBundle` has the same issue.

Upstream issue (with more details of the code path causing this): https://github.com/solana-labs/solana/issues/34693
Upstream fix to `simulateTransaction`: https://github.com/solana-labs/solana/pull/34619 

#### Summary of Changes

The changes are based on the upstream changes in `simulateTransaction`. 

Instead of directly encoding the post-account data, we call `get_encoded_account` (added in the upstream PR), which will correctly parse a token account. As this function was written for transaction post-accounts it also takes `overwrite_accounts: Option<&HashMap<Pubkey, AccountSharedData>>`, so that account state can be overridden if a transaction has changed it.

In order to provide this, we now maintain a `AccountOverrides` object in `simulateBundle`. This is the same data structure, but has some useful helper functions and I think makes sense to use here. After each transaction, this is updated (using `upsert_account_overrides`) with the `post_accounts` from the transaction. This is then passed when we encode the post-accounts for that transaction.

The result is that `simulateBundle` now correctly JSON encodes token accounts, using data from the transactions, or the bank if required.


mcintyre94:
Closing, superseded by #600

PR #600, opened 2024-03-05
Return parsed token accounts in simulateBundle

mcintyre94:
Rewrite of #592  on latest master

#### Problem

If you call `simulateBundle` and pass a token account in `pre/postExecutionAccountsConfigs`, then the token accounts are never returned in JSON. You will always get the base64 encoded data.

This is caused by attempting to encode the data without fetching the number of decimals for the mint, which causes a fallback to base64 data.

This was fixed for `simulateTransaction` upstream, but `simulateBundle` has the same issue.

Upstream issue (with more details of the code path causing this): https://github.com/solana-labs/solana/issues/34693
Upstream fix to `simulateTransaction`: https://github.com/solana-labs/solana/pull/34619 

#### Summary of Changes

The changes are based on the upstream changes in `simulateTransaction`. 

Instead of directly encoding the post-account data, we call `get_encoded_account` (added in the upstream PR), which will correctly parse a token account. As this function was written for transaction post-accounts it also takes `overwrite_accounts: Option<&HashMap<Pubkey, AccountSharedData>>`, so that account state can be overridden if a transaction has changed it.

In order to provide this, we now maintain a `AccountOverrides` object in `rpc_bundle_result_from_bank_result`. This is the same data structure, but has some useful helper functions and I think makes sense to use here. After each transaction, this is updated (using `upsert_account_overrides`) with the `post_accounts` from the transaction. This is then passed when we encode the post-accounts for that transaction.

The result is that `rpc_bundle_result_from_bank_result` (and the `simulateBundle` RPC call) now correctly JSON encodes token accounts, using data from the transactions, or the bank if required.


rpc/src/rpc.rs
```diff
@@ -3373,15 +3393,35 @@ pub mod utils {
                 let pre_execution_accounts = if let Some(pre_tx_accounts) =
                     bundle_output.pre_tx_execution_accounts().get(index)
                 {
-                    try_encode_accounts(pre_tx_accounts, account_encoding)?
+                    try_encode_accounts(
+                        pre_tx_accounts,
+                        account_encoding,
+                        bank,
+                        Some(&account_overrides),
```
buffalu, line 3400: do you need to use account overrides here? you should have access to the accounts in the execution results

mcintyre94: It comes from the post accounts, but needs to be converted to the shape `get_encoded_account` expects, which is exactly your `AccountOverrides` type. Upstream doesn't have that type, but it's the same shape.

Are you suggesting writing a version of `get_encoded_account` that takes the Jito `post_accounts: Vec<(Pubkey, AccountSharedData)>` directly? I thought it'd be a smaller change to re-use `get_encoded_account` from upstream by reshaping the `post_accounts` to match it.

I think I'd also need to write upsert logic to combine the outputs of the transactions in the bundle since we encode the accounts after each transaction and not just at the end. I thought it was better to re-use `AccountOverrides::upsert_account_overrides` instead of writing custom logic for that.

PR #609, opened 2024-04-11
Buffer bundles that exceed processing time and make the allowed processing time longer

buffalu:
#### Problem
The allowed processing time for bundles is too short and ones that exceed the max processing length (due to retrying from account locks or higher compute) are getting dropped.
This PR does the following:
- increases the max duration a bundle can execute from 10ms to 40ms
- any bundles that exceed the allowed time are buffered similar to ones that exceed the cost for a block.

Some metrics from the last ~12 hours on rates of simulated cus and land/forward rate
- >500k CUs: 49%
- >1M CUs: 49%
- <200k CUs: 70%
- <100k CUs: 81%
- <50k CUs: 86%

Running this commit on Jito1 and Jito2, we notice a decrease in execution timeouts.
Screenshot 2024-04-10 at 10 54 34 PM

cumsum execution timeouts
telegram-cloud-photo-size-1-4952065952367881245-x

otoh, we notice an increase in execution retries, which is expected as bundles are given more time to execute before getting rebuferred
Screenshot 2024-04-10 at 10 55 37 PM

### DIFF
--- a/core/src/banking_stage.rs
+++ b/core/src/banking_stage.rs
@@ -122,19 +122,19 @@ impl BankingStageStats {
             .load(Ordering::Relaxed) as u64
             + self.dropped_packets_count.load(Ordering::Relaxed) as u64
             + self
-            .dropped_duplicated_packets_count
-            .load(Ordering::Relaxed) as u64
+                .dropped_duplicated_packets_count
+                .load(Ordering::Relaxed) as u64
             + self.dropped_forward_packets_count.load(Ordering::Relaxed) as u64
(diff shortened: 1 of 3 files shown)


wphan:
Those land rate metrics are based on total requested CUs in the bundle?

Would forward to landed slot delta metrics also increase from this?

buffalu:
> Those land rate metrics are based on total requested CUs in the bundle?
> 
> Would forward to landed slot delta metrics also increase from this?

it's based on our simulated cu, not requested. i would expect forward to landed slot metrics to be pretty similar

PR #612, opened 2024-04-18
ignore if new validator too

michaelschem:
#### Problem

Error says to ignore if you're a hot-spare, but you should also ignore if you're a new validator because you're not yet on the leader schedule.

#### Summary of Changes

Add some text to the error message.

### DIFF
--- a/core/src/proxy/block_engine_stage.rs
+++ b/core/src/proxy/block_engine_stage.rs
@@ -173,10 +173,10 @@ impl BlockEngineStage {
             .await
             {
                 match e {
-                    // This error is frequent on hot spares, and the parsed string does not work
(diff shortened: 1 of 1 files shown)


core/src/proxy/block_engine_stage.rs
```diff
@@ -173,10 +173,10 @@ impl BlockEngineStage {
             .await
             {
                 match e {
-                    // This error is frequent on hot spares, and the parsed string does not work
+                    // This error is frequent on hot spares and on new validators, and the parsed string does not work
```
segfaultdoc, line 176: @michaelschem can you add context that new validator is not on leader schedule?

PR #615, opened 2024-05-22
take a snapshot of all bundle accounts before sim (#13)

segfaultdoc:
#### Problem
Solana state is constantly changing leading to potential race conditions when simulating.

#### Summary of Changes
Take a snapshot of all accounts to simulate account locks.

### DIFF
--- a/bundle/src/bundle_execution.rs
+++ b/bundle/src/bundle_execution.rs
@@ -17,8 +17,9 @@ use {
         transaction::{SanitizedTransaction, TransactionError, VersionedTransaction},
     },
     solana_svm::{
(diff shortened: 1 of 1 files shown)


bundle/src/bundle_execution.rs
```diff
@@ -264,6 +265,16 @@ pub fn load_and_execute_bundle<'a>(
                 account_overrides.upsert_account_overrides(
                     bank.get_account_overrides_for_simulation(&account_keys),
                 );
+
+                // An unfrozen bank's state is always changing. Need to take a snapshot of accounts to avoid race conditions.
```
buffalu, line 269: can you add comment - this doesn't perfeclty prevent race conditions but it does minimize them

PR #753, opened 2025-01-15
Move block_cost_limit tracking to BankingStage in preparation for SIMD-0207

buffalu:
#### Problem
Previously:
- we’d manually adjust the cost tracker block_cost_limit in bundle stage based on progress through bank ticks.
- each new bank would do CostTracker::default()
- there was some issues here, if we were processing a lot of bundles towards the end of the slot, the reservation wasn’t “un-reserved”.

#### Summary of Changes
- SIMD-0207 activates a higher CU limit at epoch boundary when the feature is added
- v2.1 propagates cost tracker parameters through banks in Bank::new_from_parent
- Instead of handling cost tracker reservation in BundleStage, we pass a callback that defines the behavior of how cost tracker handles the bank ticks to banking stage. This gets propagated down to the cost tracker


buffalu:
need to revist this - change is breaking CI. cost tracker now copied from the parent

jedleggett (approved):

segfaultdoc (approved):

PR #770, opened 2025-01-23
Fix Bind Addresses in Jito Proxy

jedleggett:
#### Problem
When creating the socket to connect to the block-engine and relayer, the default route is currently used instead of the bind address supplied via cli arg.  

#### Summary of Changes
This adds the bind address to the proxy configs and creates a new socket using that address.

Fixes #


core/src/proxy/block_engine_stage.rs
```diff
@@ -232,10 +247,19 @@ impl BlockEngineStage {
...
+
+        let mut http = hyper::client::connect::HttpConnector::new();
+        http.enforce_http(false);
+        http.set_local_address(Some(local_ip));
+
+        let auth_channel = timeout(
+            *connection_timeout,
+            backend_endpoint.connect_with_connector(http.clone()),
+        )
+        .await
```
badgerrupt0, line 260: might be nice to add some comments here to explain the http usage

core/src/proxy/block_engine_stage.rs
```diff
@@ -232,10 +247,19 @@ impl BlockEngineStage {
             "connecting to auth: {}",
             local_block_engine_config.block_engine_url
         );
-        let auth_channel = timeout(*connection_timeout, backend_endpoint.connect())
-            .await
-            .map_err(|_| ProxyError::AuthenticationConnectionTimeout)?
-            .map_err(|e| ProxyError::AuthenticationConnectionError(e.to_string()))?;
+        let local_ip: IpAddr = local_block_engine_config.bind_address;
+
```
badgerrupt0, line 251: (nit) could chain the method calls, if it is allowed- just more a little bit more concise:

```
let mut http = hyper::client::connect::HttpConnector::new()
    .enforce_http(false)
    .set_local_address(local_ip.into());
```

core/src/proxy/block_engine_stage.rs
```diff
@@ -74,13 +75,27 @@ pub struct BlockBuilderFeeInfo {
...
+    pub bind_address: IpAddr,
+}
+
+impl Default for BlockEngineConfig {
+    fn default() -> Self {
+        Self {
+            block_engine_url: String::default(),
+            trust_packets: bool::default(),
+            bind_address: IpAddr::V4(Ipv4Addr::new(0, 0, 0, 0)),
+        }
```
badgerrupt0, line 97: There's a [constant for this](https://doc.rust-lang.org/std/net/struct.Ipv4Addr.html#associatedconstant.UNSPECIFIED) if you'd prefer it. I think either works

badgerrupt0 (commented):
Looks good! Just some minor comments for style and documentation

PR #834, opened 2025-04-02
don't clone every packet...

cavemanloverboy:
#### Problem
every packet is cloned

#### Summary of Changes
every packet is no longer cloned


buffalu:
i think new master takes a reference to the packet - doing the master rebase right now

PR #844, opened 2025-04-05
Add better error handling around signatures

buffalu:
#### Problem
- Missing signatures cause failure of bundle id creation.
- Some providers likely stripping signatures to avoid having entire tx contents

#### Summary of Changes
- Re-order operations for signature verification
- If skipping sigverify, allow for bad bundle ids and use default

### DIFF
--- a/rpc/src/rpc.rs
+++ b/rpc/src/rpc.rs
@@ -4409,6 +4409,17 @@ pub mod rpc_full {
                 });
             }
 
+            // if one can't derive bundle id and we're not skipping signature verification then it's an error
+            let bundle_id = derive_bundle_id(&decoded_transactions);
+            if let Err(missing_signature_index) = &bundle_id {
+                if !config.skip_sig_verify {
+                    return Err(Error::invalid_params(format!(
+                        "Transaction index {} missing signature",
+                        missing_signature_index
+                    )));
+                }
+            }
+
             let runtime_txs = decoded_transactions
                 .into_iter()
                 .map(|tx| sanitize_transaction(tx, bank.as_ref(), bank.get_reserved_account_keys()))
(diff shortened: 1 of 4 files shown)


rpc/src/rpc.rs
```diff
@@ -4409,22 +4409,31 @@ pub mod rpc_full {
...
+                if !config.skip_sig_verify {
+                    return Err(Error::invalid_params(format!(
+                        "Transaction index {} missing signature",
+                        missing_signature_index
+                    )));
+                }
+            }
+            let sanitized_bundle = SanitizedBundle {
+                transactions: runtime_txs,
+                bundle_id: bundle_id.unwrap_or_default(),
```
esemeniuc, line 4434: is this gonna blow up later? its an empty string

buffalu: its not used anywhere except logging

buffalu: maybe i can put some string here that represents "idk" instead of empty

rpc/src/rpc.rs
```diff
@@ -4409,22 +4409,32 @@ pub mod rpc_full {
...
             if !config.skip_sig_verify {
-                for tx in &sanitized_bundle.transactions {
+                for tx in &runtime_txs {
                     verify_transaction(tx, &bank.feature_set)?;
                 }
             }
 
+            let sanitized_bundle = SanitizedBundle {
+                transactions: runtime_txs,
+                bundle_id: bundle_id.unwrap_or("unknown".to_string()),
```
buffalu, line 4435: lmk what you think of this @esemeniuc

esemeniuc: lgtm

esemeniuc (approved):

dgski (approved):

PR #888, opened 2025-06-17
Remove the tip distributor code

buffalu:
#### Problem
The tip distributor code is not used anymore and was replaced by tip router. Managing it is annoying.

#### Summary of Changes
Remove the code


ebatsell (approved):
💯

PR #900, opened 2025-06-27
Feature: Adding Priority Fee Sharing Scrip

coachchucksol:
Adds the Priority Fee Sharing Script to be installed alongside the validator.


coachchucksol:
Close in favor of #963

PR #920, opened 2025-07-09
Automatically use optimal Block Engine region

esemeniuc:

Changes:


core/src/proxy/block_engine_stage.rs
```diff
@@ -202,36 +226,88 @@ impl BlockEngineStage {
...
-            backend_endpoint = backend_endpoint
-                .tls_config(tonic::transport::ClientTlsConfig::new())
-                .map_err(|_| {
-                    ProxyError::BlockEngineConnectionError(
-                        "failed to set tls_config for block engine service".to_string(),
-                    )
-                })?;
+            Self::get_endpoint(local_block_engine_config.block_engine_url.clone())?;
+
+        if !local_block_engine_config.disable_block_engine_autoconfig {
```
esemeniuc: use new discovery stuff here

turbine/src/broadcast_stage.rs
```diff
@@ -121,7 +122,8 @@ impl BroadcastStageType {
         bank_forks: Arc<RwLock<BankForks>>,
         shred_version: u16,
         quic_endpoint_sender: AsyncSender<(SocketAddr, Bytes)>,
-        shred_receiver_address: Arc<RwLock<Option<SocketAddr>>>,
+        shredstream_receiver_address: Arc<ArcSwap<Option<SocketAddr>>>,
```
jedleggett, line 125: what is the difference between `shredstream_receiver_address` and `shred_receiver_address`?

core/src/proxy/block_engine_stage.rs
```diff
@@ -206,28 +400,6 @@ impl BlockEngineStage {
         // Get a copy of configs here in case they have changed at runtime
         let keypair = cluster_info.keypair().clone();
 
-        let mut backend_endpoint =
```
esemeniuc, line 209: moved to endpoint parameter and get_endpoint

turbine/src/broadcast_stage.rs
esemeniuc: 1 is auto configured, 1 is manually configured. we dedup if they are the same

jedleggett: makes sense

core/src/proxy/block_engine_stage.rs
```diff
@@ -143,23 +173,67 @@ impl BlockEngineStage {
...
-            };
+            let local_block_engine_config =
+                task::block_in_place(|| block_engine_config.lock().unwrap().clone());
             if !Self::is_valid_block_engine_config(&local_block_engine_config) {
-                sleep(CONNECTION_BACKOFF).await;
-            } else if let Err(e) = Self::connect_auth_and_stream(
+                sleep(Self::CONNECTION_BACKOFF).await;
+            }
+
+            if let Err(e) = Self::connect_auth_and_stream_maybe_autoconfig(
```
jedleggett, line 188: i think this should be an else if.  you don't want to connect if config is not valid.

esemeniuc: thanks, adding continue to above

jedleggett (approved):

PR #923, opened 2025-07-22
[v2.3] Automatically use optimal Block Engine region

esemeniuc:

Changes:

Port of #920, but with autoconfig default change to be reverted in near future

### DIFF
--- a/multinode-demo/bootstrap-validator.sh
+++ b/multinode-demo/bootstrap-validator.sh
@@ -112,7 +112,7 @@ while [[ -n $1 ]]; do
     elif [[ $1 == --block-engine-url ]]; then
       args+=("$1" "$2")
       shift 2
-    elif [[ $1 == --disable-block-engine-autoconfig ]]; then
+    elif [[ $1 == --enable-block-engine-autoconfig ]]; then
       args+=("$1")
       shift
     elif [[ $1 == --tip-payment-program-pubkey ]]; then
--- a/multinode-demo/validator.sh
+++ b/multinode-demo/validator.sh
@@ -100,7 +100,7 @@ while [[ -n $1 ]]; do
     elif [[ $1 == --tip-distribution-program-pubkey ]]; then
       args+=("$1" "$2")
       shift 2
-    elif [[ $1 == --disable-block-engine-autoconfig ]]; then
+    elif [[ $1 == --enable-block-engine-autoconfig ]]; then
       args+=("$1")
       shift
     elif [[ $1 == --commission-bps ]]; then
--- a/validator/src/commands/block_engine/mod.rs
+++ b/validator/src/commands/block_engine/mod.rs
@@ -15,10 +15,10 @@ pub fn command(_default_args: &DefaultArgs) -> App<'_, '_> {
                 .required(true)
         )
         .arg(
-            Arg::with_name("disable_block_engine_autoconfig")
-                .long("disable-block-engine-autoconfig")
+            Arg::with_name("enable_block_engine_autoconfig")
+                .long("enable-block-engine-autoconfig")
                 .takes_value(false)
-                .help("Disables Block Engine auto-configuration. This stops the validator client from using the most performant Block Engine region. Values provided to `--block-engine-url` will be used as-is."),
+                .help("Enables Block Engine auto-configuration. This stops the validator client from using the most performant Block Engine region. Values provided to `--block-engine-url` will be used as-is."),
         )
         .arg(
             Arg::with_name("trust_block_engine_packets")
@@ -30,8 +30,10 @@ pub fn command(_default_args: &DefaultArgs) -> App<'_, '_> {
 
 pub fn execute(subcommand_matches: &ArgMatches, ledger_path: &Path) -> Result<()> {
     let block_engine_url = value_t_or_exit!(subcommand_matches, "block_engine_url", String);
+    // Temporary default change: disable autoconfig by default until users are comfortable
+    // with it; explicit flag remains supported
     let disable_block_engine_autoconfig =
-        subcommand_matches.is_present("disable_block_engine_autoconfig");
+        !subcommand_matches.is_present("enable_block_engine_autoconfig");
     let trust_packets = subcommand_matches.is_present("trust_block_engine_packets");
     let admin_client = admin_rpc_service::connect(ledger_path);
 
--- a/validator/src/commands/run/args.rs
+++ b/validator/src/commands/run/args.rs
@@ -1694,11 +1694,11 @@ pub fn add_args<'a>(app: App<'a, 'a>, default_args: &'a DefaultArgs) -> App<'a, 
             .help("Number of CUs to allocate for bundles at beginning of slot.")
     )
     .arg(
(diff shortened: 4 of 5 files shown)


core/src/proxy/block_engine_stage.rs
```diff
@@ -170,7 +242,155 @@ impl BlockEngineStage {
...
+                if uri.host().is_none() {
+                    return None;
+                }
+                Some((endpoint, uri))
+            })
+            .collect_vec();
+        let ping_res = futures::future::join_all(
+            endpoints_to_ping
+                .iter()
+                .map(|(_endpoint, uri)| Self::ping(uri.host().unwrap())),
```
buffalu: don't unwrap

core/src/proxy/block_engine_stage.rs
```diff
@@ -170,7 +242,155 @@ impl BlockEngineStage {
...
+        let mut endpoint_discovery = BlockEngineValidatorClient::connect(backend_endpoint.clone())
+            .await
+            .map_err(|e| ProxyError::BlockEngineConnectionError(e.to_string()))?;
+        let endpoints = endpoint_discovery
+            .get_block_engine_endpoints(GetBlockEngineEndpointRequest {})
+            .await
+            .map_err(|e| ProxyError::BlockEngineConnectionError(e.to_string()))?
+            .into_inner();
+
+        const PING_COUNT: usize = 3;
```
buffalu: move to function

core/src/proxy/block_engine_stage.rs
```diff
@@ -170,7 +242,155 @@ impl BlockEngineStage {
...
+        ping_res.iter().zip(endpoints_to_ping.iter()).for_each(
+            |(maybe_ping_res, (endpoint, _uri))| {
+                let Ok(latency_us) = maybe_ping_res.as_ref() else {
+                    return;
+                };
+
+                // todo: only show best ping
+                datapoint_info!(
+                    "block_engine_stage-ping",
+                    "endpoint" => endpoint.block_engine_url,
```
buffalu: may run into issue with bad characters here?

core/src/proxy/block_engine_stage.rs
```diff
@@ -181,20 +401,26 @@ impl BlockEngineStage {
...
                         datapoint_warn!(
                             "block_engine_stage-proxy_error",
-                            ("count", error_count, i64),
+                            ("count", *error_count, i64),
                             ("error", e.to_string(), String),
                         );
                     }
                 }
-                sleep(CONNECTION_BACKOFF).await;
+                sleep(Self::CONNECTION_TIMEOUT).await;
```
buffalu: should we break out here? it seems like we'll connect to next best BE

core/src/proxy/block_engine_stage.rs
```diff
@@ -170,7 +242,155 @@ impl BlockEngineStage {
...
+                u64,        /* latency us */
+            ),
+        > = ahash::HashMap::default();
+        ping_res.iter().zip(endpoints_to_ping.iter()).for_each(
+            |(maybe_ping_res, (endpoint, _uri))| {
+                let Ok(latency_us) = maybe_ping_res.as_ref() else {
+                    return;
+                };
+
+                // todo: only show best ping
```
buffalu: remove todo, this seems fine

core/src/proxy/block_engine_stage.rs
esemeniuc: the issue with moving into the ping fn is we want to run a single ping so we don't have to wait for each individual response for the set of 3. eg you are in FRA, pinging TYO, you dont want to wait 250ms for each ping (total of 750ms), rather we want them to be fired off concurrently to avoid holding up the connection startup loop

core/src/proxy/block_engine_stage.rs
esemeniuc: its checked above. issue is rust compiler complaining about passing around string refs to functions that have already returned (lifetime issues)

core/src/proxy/block_engine_stage.rs
buffalu: yeah, mainly saying move the parallel pinging to a diff function. this one pretty beefy right now

core/src/proxy/block_engine_stage.rs
```diff
@@ -145,32 +172,224 @@ impl BlockEngineStage {
...
+        banking_packet_sender: &BankingPacketSender,
+        exit: &Arc<AtomicBool>,
+        block_builder_fee_info: &Arc<Mutex<BlockBuilderFeeInfo>>,
+        error_count: &mut u64,
+        shredstream_receiver_address: &Arc<ArcSwap<Option<SocketAddr>>>,
+    ) -> crate::proxy::Result<()> {
+        let mut endpoint_discovery = BlockEngineValidatorClient::connect(backend_endpoint.clone())
+            .await
+            .map_err(|e| ProxyError::BlockEngineConnectionError(e.to_string()))?;
+        let endpoints = endpoint_discovery
```
ebin-mathews: Curious: Do we really need an "initial" block engine connection to get this info? Shouldn't this be decoupled from block engine itself so that the validator simply fetches this info from an external server and connects to the optimal block engine?

Are we using block engine as a server here for ease of use or is there any other reason this is coupled to block engine?

Trying to see why we even need to provide an initial block engine url here if the idea is to do a self-discovery to connect to the most optimal block engine

@esemeniuc @buffalu @jedleggett

esemeniuc: i think its best to avoid fetching from external services for 2 reasons:
1. you need to connect to the block engine in the first place. don't want to have more complex setup than necessary
2. validator service to should always be working
3. how will you get the new non block engine endpoint to the validator? i don't think hard coding it is a better solution

validator/src/cli.rs
```diff
@@ -319,6 +319,12 @@ fn deprecated_arguments() -> Vec<DeprecatedArg> {
         .takes_value(true)
         .validator(is_parsable::<usize>)
         .help("The maximum size in bytes to which the outgoing websocket buffer can grow."));
+    add_arg!(Arg::with_name("shred_receiver_address")
```
buffalu: is there an easy way to not override custom non-jito shred receiver addresses?

core/src/proxy/block_engine_stage.rs
esemeniuc: added comment

core/src/proxy/block_engine_stage.rs
esemeniuc: updated

core/src/proxy/block_engine_stage.rs
esemeniuc: moved now

validator/src/cli.rs
esemeniuc: rm'd now

core/src/proxy/block_engine_stage.rs
esemeniuc: we want to connect to next best Block engine, correct.

PR #974, opened 2025-09-02
[Master] Automatically use optimal Block Engine region

esemeniuc:

Changes:

Port of #923, without the autoconfig default change

### DIFF
--- a/core/src/proxy/block_engine_stage.rs
+++ b/core/src/proxy/block_engine_stage.rs
@@ -14,12 +14,14 @@ use {
             ProxyError,
         },
     },
+    arc_swap::ArcSwap,
     crossbeam_channel::Sender,
+    itertools::Itertools,
     jito_protos::proto::{
         auth::{auth_service_client::AuthServiceClient, Token},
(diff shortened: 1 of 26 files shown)


core/src/proxy/block_engine_stage.rs
```diff
@@ -305,6 +653,7 @@ impl BlockEngineStage {
         connection_timeout: &Duration,
         keypair: Arc<Keypair>,
         cluster_info: &Arc<ClusterInfo>,
+        block_engine_url: &str,
```
jedleggett, line 702: this seems redundant, since it's already stored in local_config

jedleggett: oh, it's because that's the key in your iterator

jedleggett: hmm, there's nothing keeping this in sync with local_config.  i think that breaks the disconnect logic

esemeniuc: updated

PR #997, opened 2025-10-14
Make shredstream optional

esemeniuc:
#### Summary of Changes

Fixes #


core/src/proxy/block_engine_stage.rs
buffalu, line 417: do we need to return an error here?

esemeniuc: thats the block engine global endpoint. the proto field should never be None, and we dont do dns resolution there. the logic afterwards doesnt make sense if you don't early return

PR #1063, opened 2025-12-02
Update programs to newest tip distribution + tip payment code

buffalu:
#### Problem
The current tip distribution program doesn't work with VoteStateV4.

#### Summary of Changes
Newer programs work with the new VoteStateV4 account by parsing bytes 4..36 to grab the node pubkey

Fixes #


aoikurokawa (approved):
LGTM!

PR #1090, opened 2025-12-16
fix unstaked xdp retransmit to block engine

alexpyattaev:
#### Problem

Unstaked nodes were not allowed to forward shreds to BlockEngine if XDP was enabled

#### Summary of Changes

Allow entry into send codepath even if the only retransmit target is the block engine receiver.


esemeniuc (approved):

buffalu (approved):

alexpyattaev:
Signed the commit as required

PR #1093, opened 2025-12-29
Fix bundle account locker TOCTOU (time of check time of use) bug

buffalu:
#### Problem
Bundle account locker has a time of check time of use bug where it prematurely releases the bundle locks.

#### Summary of Changes
Moves the bundle account locker release until after try_lock_accounts.

Fixes #


core/src/banking_stage/transaction_scheduler/greedy_scheduler.rs
```diff
@@ -311,6 +310,8 @@ fn try_schedule_transaction<Tx: TransactionWithMeta>(
         }
     };
 
+    drop(l_account_locks);
```
esemeniuc, line 314: wanna leave a comment so it doesn't get moved later?

PR #1098, opened 2025-12-30
Oliver feedback

buffalu:
#### Changes
- Remove unused solana-bundle-sdk
- Simplify insert_bundle callsite by removing method and chaining the recv_timeout with a try_recv
- Simplify BundleAccountLocker/BundleStorage management when processing bundles
- Simplify bundle result aggregation
- Cache `static_instruction_limit`
- Remove unused data structures

### DIFF
--- a/core/src/bundle_stage/bundle_storage.rs
+++ b/core/src/bundle_stage/bundle_storage.rs
@@ -187,10 +187,6 @@ impl BundleStorage {
             .is_active(&agave_feature_set::static_instruction_limit::id());
         let transaction_account_lock_limit = working_bank.get_transaction_account_lock_limit();
 
-        let enable_static_instruction_limit = working_bank
-            .feature_set
-            .is_active(&agave_feature_set::static_instruction_limit::id());
-
         for (idx, packet) in batch.iter().enumerate() {
             // bundles shall contain all valid packets; checked above
             let packet_data = packet.data(..).unwrap();
@@ -204,8 +200,6 @@ impl BundleStorage {
                         root_bank,
                         working_bank,
                         enable_static_instruction_limit,
-                      
-                      
                         transaction_account_lock_limit,
                         blacklisted_accounts,
                     ) {


core/src/bundle_stage.rs
```diff
@@ -517,52 +517,30 @@ impl BundleStage {
...
         let bundle = bundle_receiver.recv_timeout(recv_timeout)?;
-        Self::insert_bundle(
-            bundle_storage,
-            bundle,
-            &root_bank,
-            &working_bank,
-            blacklisted_accounts,
-            bundle_stage_metrics,
-        );
+        for bundle in std::iter::once(bundle).chain(bundle_receiver.try_iter()) {
```
buffalu, line 522: move all insert_bundle code here + do aggregation of single + multiple results into a loop

core/src/bundle_stage.rs
```diff
@@ -672,35 +650,54 @@ impl BundleStage {
...
-                    bundle_storage,
-                    bundle_account_locker,
-                    consumer,
-                    consume_worker_metrics,
-                ) {
-                    consume_worker_metrics.update_for_consume(&output);
-                    consume_worker_metrics.set_has_data(true);
-                    bundle_stage_metrics.increment_bundles_processed(1);
+
+                let result = Self::process_bundle(bank, &bundle, consumer, consume_worker_metrics);
```
buffalu: manage bundle locks and bundle storage better. bundles should always be destroyed on ok or not retryable and retried on retryable error.

core/src/bundle_stage.rs
```diff
@@ -672,35 +650,54 @@ impl BundleStage {
...
+                        // This error is not possible from this code path as any tip processing directly calls Consumer::process_and_record_aged_transactions
+                        continue;
+                    }
                 }
             }
 
             consume_worker_metrics.maybe_report_and_reset();
         }
 
         while let Some(bundle) = bundles.pop_front() {
```
OliverNChalk: Thoughts on merging the two loops? You would change the flow slightly to be as follows:

LOOP:
```rust
loop
    while bundles.len() < WINDOW_SIZE
        pop storage; push to local bundles; if storage empty break while loop;
    
    if let Some(bundle) = bundles.pop_front()
        process 1 local bundle; if empty break unconditional loop;
```

This loop will first build up the window size (or drain the storage) and then for each iteration it will consume one local before replenishing the window.

buffalu: yes, this is way cleaner. good suggestion

PR #1115, opened 2026-01-06
Fix `RelayerStage` warning logs

grod220:
#### Problem

Despite not passing `--relayer-url`, I noticed the logs were outputting:

```
WARN solana_core::proxy::relayer_stage] can't connect to relayer. missing relayer_url.
```

While harmless, it is creating unnecessary noise. After digging deeper, it appears there is a zombie thread created that enters a while loop, checks for the relayer config, logs a warning, and sleeps. Without passing a relayer_url there isn't an exit condition it seems.

#### Summary of Changes

Refactoring to use watch channels to waken up RelayerStage when it changes versus the polling. Also downgrading the log to a debug!.

### DIFF
--- a/core/src/admin_rpc_post_init.rs
+++ b/core/src/admin_rpc_post_init.rs
@@ -88,8 +88,7 @@ pub struct AdminRpcRequestMetadataPostInit {
     pub node: Option<Arc<NodeMultihoming>>,
     pub banking_control_sender: mpsc::Sender<BankingControlMsg>,
     pub block_engine_config: Arc<Mutex<BlockEngineConfig>>,
-    pub relayer_config: Arc<Mutex<RelayerConfig>>,
-    pub relayer_config_tx: Sender<RelayerConfig>,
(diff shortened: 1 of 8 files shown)


dgski:
Appreciate the patch! Good point about this thread being needless if no relayer_url is provided.

One thing to note, the code you changed runs on process start up. The `relayer_config` can still be updated via the Admin RPC during runtime.

I think the correct thing to do would be to conditionally spawn/kill the thread based on whether a valid config is available at any given moment (`Arc<Mutex<RelayerConfig>>`).

Perhaps a simpler solution in the interim would be to remove the log?

grod220:
I worry that the conditional spawn/kill logic would be complex. What do you think of a parked thread using a watch channel? It has a negligible footprint and feels pretty simple. That said, we should demote that log to a debug regardless. Let me know what you think of the latest commit.

grod220:
@dgski could you re-review 🙏

grod220:
Closing due to lack of interest

PR #1118, opened 2026-01-09
[v3.0] Fix TVU not receiving shreds in local cluster

esemeniuc:
Issue: mainnet-beta cluster will dedupe tvu ports

Solution: don't launch genesis as mainnet-beta

### DIFF
--- a/bam-local-cluster/src/cluster_manager.rs
+++ b/bam-local-cluster/src/cluster_manager.rs
@@ -354,8 +354,10 @@ impl BamLocalCluster {
             solana_local_cluster::local_cluster::DEFAULT_MINT_LAMPORTS,
             &vote_keypairs,
             stakes,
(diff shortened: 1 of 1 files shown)


bam-local-cluster/src/cluster_manager.rs
```diff
@@ -354,7 +354,8 @@ impl BamLocalCluster {
             solana_local_cluster::local_cluster::DEFAULT_MINT_LAMPORTS,
             &vote_keypairs,
             stakes,
-            ClusterType::MainnetBeta,
+            ClusterType::Development, // don't use mainnet, since we de-dupe local tvu ip addresses
```
ebin-mathews: nit: Can you add a comment on why we don't need de-dupe for local cluster?

esemeniuc: local cluster, everything is on 127.0.0.1, so we don't want to dedupe multiple validators that share the same ip

PR #1125, opened 2026-01-14
[v3.1] Shoot down block engine connection if BAM enabled

esemeniuc:
Issue:
Block engine stage can sleep for up to 1 second before another txn/bundle is received. When a validator connects to BAM, the bam_enabled flag is set. If the backend disconnects first (due to new connection), then the validator operator sees an incorrect error.

Solution:
Check bam_enabled before processing disconnect errors

Changes:
  - Replace the boolean BAM flag with atomic (disconnected/connecting/connected) so callers can distinguish readiness from in‑flight connection attempts.
  - Avoid waiting for sleep to update bam_enabled state
  - Update the BAM manager to mark `connecting` immediately, promote to `connected` only after health/config are received, and reset on identity or URL changes to prevent stale state during reconnects.
  - Update block engine connection/streaming to short‑circuit when BAM is active and map transient errors to a BamEnabled sentinel, avoiding log spam
  - Gate scheduler, fetch fallback, and bundle buffering on the connected state only so normal paths continue during connecting and only switch once BAM is ready.

### DIFF
--- a/core/src/bam_manager.rs
+++ b/core/src/bam_manager.rs
@@ -221,7 +221,7 @@ impl BamManager {
                         // with sending an auth proof w/ the old identity
                         let timeout = std::time::Duration::from_secs(180);
                         Self::wait_for_identity_in_cluster_info(
-                            new_identity.load().as_ref().clone(),
+                            *new_identity.load().as_ref(),
                             &dependencies.cluster_info,
                             timeout,
                         );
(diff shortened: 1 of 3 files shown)


core/src/proxy/block_engine_stage.rs
```diff
@@ -826,12 +820,21 @@ impl BlockEngineStage {
             }
 
             tokio::select! {
-                maybe_msg = packet_stream.message() => {
-                    let resp = maybe_msg?.ok_or(ProxyError::GrpcStreamDisconnected)?;
+                maybe_packet = packet_stream.message() => {
+                    if bam_enabled.load(Ordering::Relaxed) {
+                        info!("bam enabled, exiting block engine stage"); // BAM may have disconnected, avoid erroneous Grpc disconnected error
```
ebin-mathews:
```suggestion
                        info!("bam enabled, exiting block engine stage"); // BAM may have connected, avoid erroneous Grpc disconnected error
```

PR #1138, opened 2026-01-20
Dynamic Client Id

dgski:
Set ClientId based on if it is actively connected to the BAM Node.


0xEdgar (approved):
wonder if we should upstream the changes that touch gossip / version libs to reduce likelihood of merge conflicts in the future. only applicable if those files are being changed a lot upstream though.

dgski:
@0xEdgar It's a `pub` for two fields... I'm not worried about that

buffalu (approved):

PR #1143, opened 2026-01-22
feat: add support for multiple shred receivers

rupansh:
#### Problem

The jito client by default only allows a single address for shred receivers right now

#### Summary of Changes

This change allows specifying multiple addresses in the shred receiver related arguments, while preserving complete backwards compatibility. This can also be backported to v3.0 easily, our fork has another branch (https://github.com/Overclock-Validator/jito-solana/tree/v3.0.14-jito-multirecv) for v3.


esemeniuc:
Cleaned up and merged in #1200. Thanks for contribution

PR #1144, opened 2026-01-23
New FetchStageManager

### LINKED ISSUE #1142: FetchStageManager does not properly handle the relayer -> BAM -> relayer TPU changes
#### Problem
If a validator runs a relayer and connects to BAM, when BAM disconnects it does not properly change the TPU back to the relayer.

#### Proposed Solution
Fix it.

dgski:
#### Problem
- Current hard to follow logic
- Bug that prevents BAM -> relayer fallback
- No tests.

#### Fix
Re-write the whole thing in a much more understandable way. Move the bam tpu setting into the fetch stage manager. Also add tests.

closes #1142

### DIFF
--- a/core/src/bam_manager.rs
+++ b/core/src/bam_manager.rs
@@ -164,10 +164,7 @@ impl BamManager {
                                 }
 
                                 if let Some(builder_config) = connection.get_latest_config() {
-                                    Self::update_tpu_config(
-                                        Some(&builder_config),
-                                        &dependencies,
-                                    );
+                                    Self::update_tpu_config(Some(&builder_config), &dependencies);
                                     Self::update_block_engine_key_and_commission(
                                         Some(&builder_config),
                                         &dependencies.block_builder_fee_info,
@@ -318,7 +315,11 @@ impl BamManager {
             return;
         };
         info!("Setting TPU={tpu}, TPU Forward={tpu_fwd} from BAM config");
-        dependencies.bam_tpu_info.write().unwrap().replace((tpu, tpu_fwd));
+        dependencies
+            .bam_tpu_info
+            .write()
+            .unwrap()
+            .replace((tpu, tpu_fwd));
     }
 
     fn update_block_engine_key_and_commission(
--- a/core/src/bam_manager.rs
+++ b/core/src/bam_manager.rs
@@ -317,9 +317,7 @@ impl BamManager {
         info!("Setting TPU={tpu}, TPU Forward={tpu_fwd} from BAM config");
         dependencies
             .bam_tpu_info
-            .write()
-            .unwrap()
-            .replace((tpu, tpu_fwd));
+            .store(Arc::new(Some((tpu, tpu_fwd))));
     }
 
     fn update_block_engine_key_and_commission(
--- a/core/src/bam_manager.rs
+++ b/core/src/bam_manager.rs
@@ -166,7 +166,7 @@ impl BamManager {
                                 if let Some(builder_config) = connection.get_latest_config() {
                                     Self::update_tpu_config(
                                         Some(&builder_config),
-                                        &dependencies.cluster_info,
+                                        &dependencies,
                                     );
                                     Self::update_block_engine_key_and_commission(
                                         Some(&builder_config),
@@ -247,7 +247,7 @@ impl BamManager {
                 // Check if block builder info has changed
                 if let Some(builder_config) = connection.get_latest_config() {
                     if Some(&builder_config) != cached_builder_config.as_ref() {
-                        Self::update_tpu_config(Some(&builder_config), &dependencies.cluster_info);
+                        Self::update_tpu_config(Some(&builder_config), &dependencies);
                         Self::update_block_engine_key_and_commission(
                             Some(&builder_config),
                             &dependencies.block_builder_fee_info,
@@ -306,19 +306,19 @@ impl BamManager {
         )))
     }
 
-    fn update_tpu_config(config: Option<&ConfigResponse>, cluster_info: &Arc<ClusterInfo>) {
+    fn update_tpu_config(config: Option<&ConfigResponse>, dependencies: &BamDependencies) {
         let Some(tpu_info) = config.and_then(|c| c.bam_config.as_ref()) else {
             return;
         };
 
-        if let Some(tpu) = Self::get_sockaddr(tpu_info.tpu_sock.as_ref()) {
-            info!("Setting TPU: {tpu:?}");
-            let _ = cluster_info.set_tpu_quic(tpu);
-        }
-        if let Some(tpu_fwd) = Self::get_sockaddr(tpu_info.tpu_fwd_sock.as_ref()) {
-            info!("Setting TPU forward: {tpu_fwd:?}");
-            let _ = cluster_info.set_tpu_forwards_quic(tpu_fwd);
-        }
+        let Some(tpu) = Self::get_sockaddr(tpu_info.tpu_sock.as_ref()) else {
+            return;
+        };
+        let Some(tpu_fwd) = Self::get_sockaddr(tpu_info.tpu_fwd_sock.as_ref()) else {
+            return;
+        };
+        info!("Setting TPU={tpu}, TPU Forward={tpu_fwd} from BAM config");
+        dependencies.bam_tpu_info.write().unwrap().replace((tpu, tpu_fwd));
     }
 
     fn update_block_engine_key_and_commission(
--- a/core/src/proxy/fetch_stage_manager.rs
+++ b/core/src/proxy/fetch_stage_manager.rs
@@ -493,9 +493,9 @@ mod tests {
 
         TestContext {
             cluster_info,
-            bam_enabled: bam_enabled,
+            bam_enabled,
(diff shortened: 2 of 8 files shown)


core/src/proxy/fetch_stage_manager.rs
```diff
@@ -80,175 +40,674 @@ pub struct FetchStageManager {
...
-        heartbeat_rx: Receiver<HeartbeatEvent>,
-        // Channel that packets from FetchStage are intercepted from.
+        relayer_heartbeat_rx: Receiver<HeartbeatEvent>,
         packet_intercept_rx: Receiver<PacketBatch>,
         // Intercepted packets get piped through here.
         packet_tx: Sender<PacketBatch>,
         exit: Arc<AtomicBool>,
         bam_enabled: Arc<AtomicU8>,
         my_fallback_contact_info: contact_info::ContactInfo,
+        bam_tpu_info: Arc<RwLock<Option<(SocketAddr, SocketAddr)>>>,
```
esemeniuc: can replace rwlock -> arcswap

core/src/proxy/fetch_stage_manager.rs
```diff
@@ -80,175 +40,674 @@ pub struct FetchStageManager {
...
+    metrics: FetchStageMetrics,
+
+    /// How often to check for relayer heartbeat timeouts
+    max_time_between_relayer_heartbeats: Duration,
+
+    /// Relayer TPU enable delay
+    relayer_tpu_enable_delay: Duration,
+}
+
+impl FetchStageBrain {
```
esemeniuc: Brain? think `FetchStageTpuStateMachine` is more descriptive

core/src/proxy/fetch_stage_manager.rs
```diff
@@ -80,175 +40,674 @@ pub struct FetchStageManager {
...
+                // Run the semi-eternal loop
+                while !exit.load(Ordering::Relaxed) {
+                    let all_good = select! {
+                        recv(packet_intercept_rx) -> pkt => brain.handle_packet_batch(pkt),
+                        recv(evaluation_tick) -> _ => brain.handle_evaluation_tick(),
+                        recv(relayer_heartbeat_rx) -> tpu_info => brain.handle_relayer_message(tpu_info),
+                        recv(metrics_tick) -> _ => brain.handle_metrics_tick(),
+                    };
+                    if !all_good {
+                        break;
```
esemeniuc, line 119: should have datapoint for error

esemeniuc, line 119: should we break? continue seems safer if relayer info doesn't have good tpu address. haven't checked the parsing, but something to consider

core/src/proxy/fetch_stage_manager.rs
```diff
@@ -80,175 +40,674 @@ pub struct FetchStageManager {
...
+            relayer_info.last_heartbeat = Instant::now();
+            relayer_info.tpu_addresses.tpu_addr = tpu_addr;
+            relayer_info.tpu_addresses.tpu_forward_addr = tpu_forward_addr;
+        } else {
+            self.relayer_info = Some(HeartbeatingRelayerInfo {
+                tpu_addresses: TpuAddresses {
+                    tpu_addr,
+                    tpu_forward_addr,
+                },
+                first_heartbeat: Instant::now(),
```
esemeniuc: reuse the Instant::now()

core/src/proxy/fetch_stage_manager.rs
```diff
@@ -80,175 +40,674 @@ pub struct FetchStageManager {
...
+        TpuState {
+            tpu_type: TpuConnectionType::Original,
+            addr: self.original_tpu_info.tpu_addr,
+            fwd_addr: self.original_tpu_info.tpu_forward_addr,
+        }
+    }
+
+    /// Evaluate the current state and make transitions as needed; returns false if we should shut down
+    fn handle_evaluation_tick(&mut self) -> bool {
+        // Increment the state machine using latest gathered data
```
esemeniuc, line 281: think we need to check if heartbeat is stale:
```
        // Reset relayer tracking after a heartbeat gap so the enable-delay restarts.
        if self.relayer_info.as_ref().is_some_and(|info| {
            info.last_heartbeat.elapsed() > self.max_time_between_relayer_heartbeats
        }) {
            self.relayer_info = None;
        }
```

core/src/proxy/fetch_stage_manager.rs
esemeniuc: cant starve process or deadlock by accident

core/src/proxy/fetch_stage_manager.rs
```diff
@@ -80,175 +40,674 @@ pub struct FetchStageManager {
...
-                    state.reset_to_bam_state();
-                    // Drain any queued packets to prevent buildup
-                    while packet_intercept_rx.try_recv().is_ok() {}
-                    std::thread::sleep(Duration::from_millis(100));
-                    continue;
-                }
+        Builder::new()
+            .name("fetch-stage-manager".into())
+            .spawn(move || {
+                // Save original TPU info
```
buffalu, line 81: can you add comment here on why we do UDP (and not QUIC)? There's a quirk that the previous code mentioned.

core/src/proxy/fetch_stage_manager.rs
dgski: done

core/src/proxy/fetch_stage_manager.rs
dgski: Done 😭

core/src/proxy/fetch_stage_manager.rs
dgski: We only break if the channel disconnected

core/src/proxy/fetch_stage_manager.rs
dgski: Done

core/src/proxy/fetch_stage_manager.rs
dgski: This is done in `get_next_tpu_state`... but could theoretically be done here to clear it; I don't think it's super necessary and adds complexity.

core/src/proxy/fetch_stage_manager.rs
dgski: Done

core/src/proxy/fetch_stage_manager.rs
dgski: Done

ebin-mathews (dismissed):

core/src/proxy/fetch_stage_manager.rs
```diff
@@ -80,175 +41,678 @@ pub struct FetchStageManager {
...
+                // Run the semi-eternal loop
+                while !exit.load(Ordering::Relaxed) {
+                    let all_good = select! {
+                        recv(packet_intercept_rx) -> pkt => brain.handle_packet_batch(pkt),
+                        recv(evaluation_tick) -> _ => brain.handle_evaluation_tick(),
+                        recv(relayer_heartbeat_rx) -> tpu_info => brain.handle_relayer_message(tpu_info),
+                        recv(metrics_tick) -> _ => brain.handle_metrics_tick(),
+                    };
+                    if !all_good {
+                        datapoint_warn!("fetch-stage-manager-shutdown", ("stop", 1, i64));
```
esemeniuc:
```suggestion
                        datapoint_warn!("fetch_stage_manager-shutdown", ("stop", 1, i64));
```

esemeniuc (dismissed):

esemeniuc (approved):

core/src/proxy/fetch_stage_manager.rs
```diff
@@ -80,175 +41,678 @@ pub struct FetchStageManager {
...
+
+    fn is_bam_connected(&self) -> bool {
+        BamConnectionState::from_u8(self.bam_enabled.load(Ordering::Relaxed))
+            == BamConnectionState::Connected
+    }
+
+    fn get_next_tpu_state(&self) -> TpuState {
+        // BAM has highest priority (non-sync check is fine because downstream checks `bam_enabled` as well)
+        if self.is_bam_connected() {
+            if let Some((addr, fwd_addr)) = self.bam_tpu_info.load().as_ref() {
```
segfaultdoc, line 244: This is a weird case, make sure no race conditions here that get us into a disgusting loop

core/src/bam_manager.rs
```diff
@@ -324,19 +321,21 @@ impl BamManager {
...
-            info!("Setting TPU forward: {tpu_fwd:?}");
-            let _ = cluster_info.set_tpu_forwards_quic(tpu_fwd);
-        }
+        let Some(tpu) = Self::get_sockaddr(tpu_info.tpu_sock.as_ref()) else {
+            return;
+        };
+        let Some(tpu_fwd) = Self::get_sockaddr(tpu_info.tpu_fwd_sock.as_ref()) else {
+            return;
+        };
+        info!("Setting TPU={tpu}, TPU Forward={tpu_fwd} from BAM config");
```
0xEdgar, line 335: add datapoint for current TPU state for observability?

PR #1145, opened 2026-01-23
Simple ping pong response implementation

ebin-mathews:
#### Problem
Implement pong response from the validator

#### Summary of Changes
This PR implements a simple ping-pong response mechanism for the BAM (Block Auction Market) connection system. The changes add handling for incoming ping messages by responding with pong messages, along with corresponding metrics tracking.

Changes:

Added ping message handling in the outbound message processing loop
Introduced three new metrics to track ping-related activity (ping_sent, outbound_ping_sent, outbound_ping_fail)
Updated the bam-protos subproject commit reference

Fixes #

### DIFF
--- a/core/src/bam_connection.rs
+++ b/core/src/bam_connection.rs
@@ -10,14 +10,14 @@ use {
             ConfigRequest, ConfigResponse, SchedulerMessage, SchedulerMessageV0, SchedulerResponse,
             SchedulerResponseV0,
         },
-        bam_types::{AtomicTxnBatch, AuthProof, ValidatorHeartBeat},
+        bam_types::{AtomicTxnBatch, AuthProof, Pong, ValidatorHeartBeat},
     },
     solana_gossip::cluster_info::ClusterInfo,
     solana_keypair::Keypair,
     solana_signer::Signer,
     std::{
         sync::{
-            atomic::{AtomicBool, AtomicU64, Ordering::Relaxed},
+            atomic::{AtomicBool, AtomicU32, AtomicU64, Ordering::Relaxed},
             Arc, Mutex,
         },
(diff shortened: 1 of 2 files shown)


core/src/bam_connection.rs
```diff
@@ -478,6 +496,17 @@ impl BamConnectionMetrics {
                 self.outbound_fail.swap(0, Relaxed) as i64,
                 i64
             ),
+            ("ping_sent", self.ping_sent.swap(0, Relaxed) as i64, i64),
+            (
+                "outbound_ping_sent",
```
0xEdgar: nit: outbound_ping_sent_count? outbound_ping_fail_count?

buffalu:
lgtm after CI is fixed, something is wrong with it

core/src/bam_connection.rs
```diff
@@ -309,6 +309,19 @@ impl BamConnection {
                                     Self::send_batch_results(&mut outbound_sender, std::mem::take(&mut waiting_results), metrics.as_ref());
                                 }
                             }
+                            BamOutboundMessage::Ping(id) => {
+                                metrics.ping_received.fetch_add(1, Relaxed);
+                                metrics.last_ping_id_received.store(id, Relaxed);
+                                let pong_message_outbound = SchedulerMessageV0 {
+                                    msg: Some(Msg::Pong(Pong { id })),
+                                };
+                                if outbound_sender.try_send(v0_to_versioned_proto(pong_message_outbound)).is_err() {
```
0xEdgar, line 318: should we handle the full case explicitly here?

ebin-mathews: This should be fine as the node sends periodically. It will behave similar to leader state

PR #1159, opened 2026-02-02
[v3.1] Reduce BAM Scheduler allocs, remove lock contention in consume_work

esemeniuc:
Tldr: should reduce heap allocs from 19 -> 13 per bundle (post prost decode)
Changes:
  - Switch shared block builder fee info to ArcSwap across BAM/proxy/bundle/TPU/bench, with atomic load/store updates and tighter validation (commission bounds, fallback to previous pubkey on parse errors).
  - Reduce allocation/CPU overhead across schedulers, bundle handling, RPC, and simulation paths (SmallVec/preallocations, buffer reuse, avoid temp lock vectors, single‑pass bundle decode/dup checks, precomputed limits, single‑pass bundle packet counts).

### DIFF
--- a/core/src/bundle_stage/bundle_account_locker.rs
+++ b/core/src/bundle_stage/bundle_account_locker.rs
@@ -48,9 +48,9 @@ impl BundleAccountLocks {
         &self.write_locks
     }
 
-    pub fn lock_accounts<'a>(
+    pub fn lock_accounts<'a, T: Iterator<Item = (&'a Pubkey, bool)>, I: IntoIterator<Item = T>>(
         &mut self,
-        transaction_locks: Vec<impl Iterator<Item = (&'a Pubkey, bool)>>,
+        transaction_locks: I,
     ) {
         for transaction_lock in transaction_locks {
             for (acc, writable) in transaction_lock {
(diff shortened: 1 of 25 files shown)


rpc/src/rpc.rs
```diff
@@ -4218,21 +4218,18 @@ pub mod rpc_full {
...
-            let mut unsanitized_txs = rpc_bundle_request
-                .encoded_transactions
-                .into_iter()
-                .map(|encoded_tx| {
+            let mut packet_hashes =
+                HashSet::with_capacity(rpc_bundle_request.encoded_transactions.len());
+            let mut unsanitized_txs =
+                Vec::with_capacity(rpc_bundle_request.encoded_transactions.len());
+            for encoded_tx in rpc_bundle_request.encoded_transactions {
+                let tx =
```
esemeniuc, line 4226: this exits early when there's decode error. no sense in continuing

core/src/bundle_stage/bundle_account_locker.rs
```diff
@@ -110,12 +112,14 @@ impl BundleAccountLocker {
         transactions: &[Tx],
         bank: &Bank,
     ) -> BundleAccountLockerResult<()> {
-        let transaction_locks = Self::get_transaction_locks(transactions, bank)?;
+        let transaction_account_lock_limit = bank.get_transaction_account_lock_limit();
```
esemeniuc, line 117: we avoid allocs

runtime/src/bank.rs
```diff
@@ -3446,6 +3444,7 @@ impl Bank {
                 }
             }
 
+            let is_execution_result_err = result.is_err();
```
segfaultdoc, line 3447: why boss?

esemeniuc: we move the result, cant access the is_err() method after its moved

segfaultdoc:
fix ci rq

PR #1161, opened 2026-02-02
[master] Reduce BAM Scheduler allocs, remove lock contention in consume_work

esemeniuc:
Tldr: should reduce heap allocs from 19 -> 13 per bundle (post prost decode)
Changes:
  - Switch shared block builder fee info to ArcSwap across BAM/proxy/bundle/TPU/bench, with atomic load/store updates and tighter validation (commission bounds, fallback to previous pubkey on parse errors).
  - Reduce allocation/CPU overhead across schedulers, bundle handling, RPC, and simulation paths (SmallVec/preallocations, buffer reuse, avoid temp lock vectors, single‑pass bundle decode/dup checks, precomputed limits, single‑pass bundle packet counts).

### DIFF
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -134,8 +134,8 @@ solana-measure = { workspace = true }
 solana-message = { workspace = true }
 solana-metrics = { workspace = true }
 solana-native-token = { workspace = true }
-solana-nohash-hasher = { workspace = true }
 solana-net-utils = { workspace = true, features = ["agave-unstable-api"] }
+solana-nohash-hasher = { workspace = true }
 solana-nonce = { workspace = true }
 solana-nonce-account = { workspace = true }
 solana-packet = { workspace = true }


core/src/banking_stage/transaction_scheduler/bam_scheduler.rs
```diff
@@ -163,12 +164,11 @@ impl<Tx: TransactionWithMeta> BamScheduler<Tx> {
             let txns = batch_ids
                 .iter()
                 .filter_map(|txn_id| container.get_transaction(*txn_id))
-                .collect_vec();
+                .collect::<SmallVec<[&Tx; 5]>>();
```
ebin-mathews: Where is 5 from? This is a whole batch and not a single bundle right?

core/src/banking_stage/transaction_scheduler/bam_scheduler.rs
```diff
@@ -163,12 +164,11 @@ impl<Tx: TransactionWithMeta> BamScheduler<Tx> {
...
 
             if self.extra_checks_enabled {
-                let lock_results = (0..txns.len())
-                    .map(|_| Ok(()))
-                    .collect::<Vec<solana_transaction_error::TransactionResult<()>>>();
+                let lock_results: SmallVec<[solana_transaction_error::TransactionResult<()>; 5]> =
```
ebin-mathews: I think this is also not bundle specific, probably not doing much here?

core/src/banking_stage/transaction_scheduler/bam_scheduler.rs
```diff
@@ -180,6 +180,7 @@ impl<Tx: TransactionWithMeta> BamScheduler<Tx> {
                     .find_position(|res| res.is_err())
                     .map(|(i, res)| (i, res.as_ref().err().unwrap().clone()))
                 {
+                    drop(txns);
```
ebin-mathews, line 189: Necessary?

core/src/banking_stage/transaction_scheduler/bam_scheduler.rs
```diff
@@ -261,6 +264,7 @@ impl<Tx: TransactionWithMeta> BamScheduler<Tx> {
                     .find_position(|res| res.is_err())
                     .map(|(i, res)| (i, res.as_ref().err().unwrap().clone()))
                 {
+                    drop(sanitized_txs);
```
ebin-mathews, line 273: Drops to keep compiler happy due to the borrow?

core/src/banking_stage/transaction_scheduler/bam_scheduler.rs
```diff
@@ -283,13 +287,13 @@ impl<Tx: TransactionWithMeta> BamScheduler<Tx> {
...
             Self::generate_work(&mut work, batch_id, &[id], revert_on_error, container, slot);
-            self.send_to_worker(vec![id], work, slot);
+            self.send_to_worker(SmallVec::from([id]), work, slot);
         }
     }
 
     fn send_to_worker(
         &mut self,
-        priority_ids: Vec<TransactionPriorityId>,
+        priority_ids: SmallVec<[TransactionPriorityId; 1]>,
```
ebin-mathews, line 303: Could be more than 1? maybe add reasoning for using 1

core/src/banking_stage/consumer.rs
```diff
@@ -526,18 +512,16 @@ impl Consumer {
...
-                .zip(batch.sanitized_transactions())
-                .filter_map(|(processing_result, tx)| {
-                    if processing_result.was_processed() {
-                        Some((tx.to_versioned_transaction(), tx))
-                    } else {
-                        None
-                    }
-                })
-                .collect_vec());
+        let processed_transactions = processing_results
```
ebin-mathews, line 515: missing the measure_us?

core/src/banking_stage/consumer.rs
```diff
@@ -547,9 +531,8 @@ impl Consumer {
         // Entries do **not** yet support conflicting transactions. To get around this we create
         // lists of transactions that are non-conflicting to shred out into entries. If we don't do
         // this, then blocks are rejected by consensus/replay.
-        let batches = Self::create_sequential_non_conflicting_batches(
-            &mut reusables,
-            processed_transactions.into_iter(),
+        let (batches, processing_results_to_transactions_us) = measure_us!(
```
ebin-mathews: maybe need a new metric here, processing_results_to_transactions_us was used in a different place

core/src/banking_stage/consumer.rs
```diff
@@ -691,7 +674,9 @@ impl Consumer {
             }
 
             if has_contention {
-                result.push(std::mem::take(&mut current_batch));
+                let mut finished_batch = Vec::with_capacity(current_batch.len());
```
ebin-mathews: Wasn't 
```
result.push(std::mem::take(&mut current_batch));
```
better?

core/src/banking_stage/consumer.rs
```diff
@@ -701,7 +686,9 @@ impl Consumer {
         }
 
         if !current_batch.is_empty() {
-            result.push(current_batch);
+            let mut finished_batch = Vec::with_capacity(current_batch.len());
+            finished_batch.append(&mut current_batch);
+            result.push(finished_batch);
```
ebin-mathews: why not simply result.push(current_batch);

core/src/bundle_stage/bundle_account_locker.rs
```diff
@@ -48,9 +48,9 @@ impl BundleAccountLocks {
         &self.write_locks
     }
 
-    pub fn lock_accounts<'a>(
+    pub fn lock_accounts<'a, T: Iterator<Item = (&'a Pubkey, bool)>, I: IntoIterator<Item = T>>(
```
ebin-mathews, line 51: style diff?

core/src/bundle_stage/bundle_account_locker.rs
```diff
@@ -125,43 +131,27 @@ impl BundleAccountLocker {
...
+            transactions
+                .iter()
+                .map(|tx| TransactionAccountLocksIterator::new(tx).accounts_with_is_writable()),
+        );
         Ok(())
     }
 
-    /// Returns the read and write locks for this bundle
-    /// Each lock type contains a HashMap which maps Pubkey to number of locks held
-    fn get_transaction_locks<'a, Tx: TransactionWithMeta>(
```
ebin-mathews, line 139: Any way to re-use this? 
Seems much simpler + testable

core/src/proxy/block_engine_stage.rs
```diff
@@ -917,13 +917,11 @@ impl BlockEngineStage {
...
-                    {
-                        let mut bb_fee = block_builder_fee_info.lock().unwrap();
-                        bb_fee.block_builder_commission = block_builder_info.commission;
-                        if let Ok(pk) = Pubkey::from_str(&block_builder_info.pubkey) {
-                            bb_fee.block_builder = pk
-                        }
-                    }
+                    block_builder_fee_info.store(Arc::new(BlockBuilderFeeInfo {
+                        block_builder: Pubkey::from_str(&block_builder_info.pubkey)
+                            .unwrap_or_else(|_| block_builder_fee_info.load().block_builder),
```
ebin-mathews: datapoint/log on fail?

core/src/banking_stage/transaction_scheduler/bam_scheduler.rs
esemeniuc:
```
error[E0502]: cannot borrow `*container` as mutable because it is also borrowed as immutable
   --> core/src/banking_stage/transaction_scheduler/bam_scheduler.rs:183:21
    |
166 |                 .filter_map(|txn_id| container.get_transaction(*txn_id))
    |                             -------- --------- first borrow occurs due to use of `*container` in closure
    |                             |
    |                             immutable borrow occurs here
...
183 |                     container.remove_by_id(next_batch_id.id);
    |                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ mutable borrow occurs here
...
205 |         }
    |         - immutable borrow might be used here, when `txns` is dropped and runs the `Drop` code for type `SmallVec`
```

core/src/proxy/block_engine_stage.rs
```diff
@@ -936,32 +934,29 @@ impl BlockEngineStage {
...
-                        bundle
-                            .bundle?
-                            .packets
-                            .into_iter()
-                            .map(proto_packet_to_packet)
-                            .collect::<Vec<BytesPacket>>(),
-                    ),
-                    bundle.uuid,
-                ))
+                let packet_batch = PacketBatch::from(
```
ebin-mathews, line 960: Refactor? Seems to be doing the same thing

core/src/banking_stage/transaction_scheduler/bam_scheduler.rs
esemeniuc: same as above

core/src/banking_stage/transaction_scheduler/bam_scheduler.rs
esemeniuc: the only use case is ever size 1, see https://github.com/jito-foundation/jito-solana/blob/d302620e2537e21441e70ddd5dc5a6d3cc8f64d3/core/src/banking_stage/transaction_scheduler/bam_scheduler.rs#L290

core/src/bundle_stage.rs
```diff
@@ -824,13 +826,12 @@ impl BundleStage {
...
-                sanitized_epoch: bank.epoch(),
-                alt_invalidation_slot: bank.slot(),
-            };
-            crank_tip_program_transactions.len()
-        ];
+        let max_age = MaxAge {
+            sanitized_epoch: bank.epoch(),
+            alt_invalidation_slot: bank.slot(),
+        };
+        let max_ages: SmallVec<[MaxAge; 2]> =
```
ebin-mathews, line 835: Why 2 for smallvec size?

core/src/tip_manager.rs
```diff
@@ -488,8 +489,8 @@ impl TipManager {
         bank: &Bank,
         keypair: &Keypair,
         block_builder_fee_info: &BlockBuilderFeeInfo,
-    ) -> Result<Vec<RuntimeTransaction<SanitizedTransaction>>> {
-        let mut transactions = Vec::new();
+    ) -> Result<SmallVec<[RuntimeTransaction<SanitizedTransaction>; 2]>> {
+        let mut transactions = SmallVec::with_capacity(2);
```
ebin-mathews, line 495: Reason for magic number 2?

core/src/banking_stage/consumer.rs
esemeniuc: its a stream, measuring time not very useful here since the caller is the one to spend the time, hence the time measure_us being moved down

core/src/banking_stage/consumer.rs
esemeniuc: its execution time is now spent in this function

core/src/bundle_stage/bundle_account_locker.rs
esemeniuc: this is for using an iter, rather than requiring us to collect

core/src/proxy/block_engine_stage.rs
esemeniuc: this is now a single pass over the bundles, vs 2 pass (before)

core/src/bundle_stage.rs
esemeniuc: see `get_initialize_tip_programs_bundle`. it only returns up to 2 elements

core/src/tip_manager.rs
esemeniuc: see `get_tip_programs_crank_bundle`, we only ever push upto twice

core/src/proxy/block_engine_stage.rs
esemeniuc: can add. old behaviour was worse (sending tips to default pubkey)

core/src/banking_stage/transaction_scheduler/bam_scheduler.rs
esemeniuc: its from get_batch, see the return type of https://github.com/jito-foundation/jito-solana/blob/d302620e2537e21441e70ddd5dc5a6d3cc8f64d3/core/src/banking_stage/transaction_scheduler/transaction_state_container.rs#L88

core/src/banking_stage/transaction_scheduler/bam_scheduler.rs
esemeniuc: this must be the same size as `txns` on line 164. 5 is just the max size before we heap allocate. the actual len is still `txns.len()`

core/src/banking_stage/transaction_scheduler/bam_scheduler.rs
ebin-mathews: question: why did you choose 5?

core/src/banking_stage/transaction_scheduler/bam_scheduler.rs
ebin-mathews: question: why did you choose 5 here?

core/src/banking_stage/transaction_scheduler/bam_scheduler.rs
ebin-mathews: oh, interesting wonder why we even have Vec here then.

core/src/banking_stage/consumer.rs
ebin-mathews: Should probably use a new metric imo and not use an older one imo

core/src/banking_stage/consumer.rs
ebin-mathews: should probably use a new metric imo

core/src/proxy/block_engine_stage.rs
ebin-mathews: where Is the 2nd pass? 😕

core/src/banking_stage/transaction_scheduler/bam_scheduler.rs
esemeniuc: i'm not sure, can remove later i guess

core/src/bundle_stage.rs
ebin-mathews: maybe add a comment regarding the magic numbers?

core/src/banking_stage/consumer.rs
esemeniuc: updated to have new metric

core/src/tip_manager.rs
ebin-mathews: request: add a comment stating the same

core/src/proxy/block_engine_stage.rs
esemeniuc:
```
 block_engine_stats.num_bundle_packets.add_assign(
            bundles
                .iter()
                .map(|bundle| bundle.batch().len() as u64)
                .sum::<u64>(),
```

core/src/bundle_stage/bundle_account_locker.rs
esemeniuc: what do you want to reuse with? new version doesnt require allocs

core/src/banking_stage/consumer.rs
esemeniuc: there's the constant `TARGET_NUM_TRANSACTIONS_PER_BATCH = 64`. when we do `take`, it replaces the vec with an empty one that has no capacity. the subsequent iteration will have a bunch of extra re-allocs due to 0 capacity vec

core/src/banking_stage/consumer.rs
esemeniuc: good call, deleted

core/src/banking_stage/transaction_scheduler/bam_scheduler.rs
esemeniuc: i didnt choose 5, david did

esemeniuc: i'm following his size

core/src/banking_stage/transaction_scheduler/bam_scheduler.rs
esemeniuc: ditto from above

core/src/banking_stage/consumer.rs
ebin-mathews: maybe this then?

```
                result.push(std::mem::replace(&mut current_batch, Vec::with_capacity(TARGET_NUM_TRANSACTIONS_PER_BATCH)));
```

core/src/proxy/block_engine_stage.rs
esemeniuc: added

core/src/bundle_stage/bundle_account_locker.rs
ebin-mathews: The deleted code is easily testable wrt  get_transaction_locks and getting the locks. The new setup don't seem to have a similar testability, hence the comment.


Is it possible to refactor so that the get_transaction_locks returns an iterator instead?

core/src/bundle_stage.rs
esemeniuc: added

core/src/tip_manager.rs
esemeniuc: added

core/src/banking_stage/consumer.rs
esemeniuc: updated

core/src/bundle_stage/bundle_account_locker.rs
esemeniuc: updated

core/src/banking_stage/transaction_scheduler/bam_receive_and_buffer.rs
```diff
@@ -75,10 +76,13 @@ pub struct BamReceiveAndBuffer {
...
-    pub txns_max_age: Vec<(
-        RuntimeTransaction<ResolvedTransactionView<SharedBytes>>,
-        MaxAge,
-    )>,
+    // SmallVec 5: BAM bundles are capped at 5 transactions (BundleStorage::MAX_PACKETS_PER_BUNDLE).
+    pub txns_max_age: SmallVec<
+        [(
+            RuntimeTransaction<ResolvedTransactionView<SharedBytes>>,
+            MaxAge,
+        ); 5],
```
buffalu: we should have a variable for this somewhere, lets use that to avoid a panic from something too big

buffalu: oh wait this is small vec, not array vec. brain fart, ignore. shouuld still use a constant

esemeniuc: the import is weird, hence the comment on line 79

core/src/banking_stage/transaction_scheduler/bam_receive_and_buffer.rs
```diff
@@ -310,25 +315,24 @@ impl BamReceiveAndBuffer {
         let enable_static_instruction_limit = root_bank
             .feature_set
             .is_active(&agave_feature_set::static_instruction_limit::ID);
-
         let mut cost: u64 = 0;
-        let mut txns_max_age = Vec::with_capacity(verified_batch.len());
+        let mut txns_max_age = SmallVec::with_capacity(verified_batch.len());
+
+        if vote_only {
```
esemeniuc, line 321: only need to check this once

core/src/banking_stage/consumer.rs
```diff
@@ -271,25 +264,23 @@ impl Consumer {
...
-                                let is_writable_conflict = writable
-                                    && (l_bundle_account_locker.write_locks().contains_key(acc)
-                                        || l_bundle_account_locker.read_locks().contains_key(acc));
-                                let is_read_conflict = !writable
-                                    && l_bundle_account_locker.write_locks().contains_key(acc);
-
-                                if is_writable_conflict || is_read_conflict {
-                                    return Err(TransactionError::AccountInUse);
-                                }
+                        let Some(l_bundle_account_locker) = &l_bundle_account_locker else {
```
esemeniuc, line 267: less indentation

core/src/banking_stage/consumer.rs
```diff
@@ -475,14 +461,14 @@ impl Consumer {
...
+                    .processing_results
+                    .iter()
+                    .map(|result| {
                         CommitTransactionDetails::NotCommitted(
-                            transaction_errors
-                                .get(index)
-                                .map(|error| {
-                                    error.clone().unwrap_or(TransactionError::CommitCancelled)
-                                })
+                            result
```
esemeniuc, line 469: skip cloning error

core/src/banking_stage/consumer.rs
esemeniuc, line 515: convert to streaming, measure_us doesnt make sense

core/src/bundle_stage/bundle_account_locker.rs
```diff
@@ -139,29 +141,19 @@ impl BundleAccountLocker {
...
-                } else {
-                    None
-                }
-            })
-            .collect::<Vec<_>>();
-
-        if transaction_locks.len() != transactions.len() {
-            return Err(BundleAccountLockerError::LockingError);
+    ) -> BundleAccountLockerResult<
+        impl Iterator<Item = impl Iterator<Item = (&'a Pubkey, bool)> + use<'a, Tx>> + use<'a, Tx>,
```
esemeniuc, line 145: converts to stream, no collect

core/src/banking_stage/transaction_scheduler/bam_scheduler.rs
ebin-mathews: nit: 5 might simply be the max number of transactions in a bundle? Can we use that const here? or define a new one?
Better to not have these magic numbers

core/src/bundle_stage/bundle_account_locker.rs
ebin-mathews: Thank you ❤️

ebin-mathews (dismissed):

ebin-mathews (approved):

PR #1172, opened 2026-02-06
Seg/fix fetch stage manager

segfaultdoc:
#### Problem
Buggy fetch stage manager when switching between relayer and BAM node. This was affecting validator's gossip TPU.

#### Summary of Changes
Centralize and clean up managing gossip info of TPU.


core/src/proxy/fetch_stage_manager.rs
```diff
@@ -82,173 +44,702 @@ impl FetchStageManager {
...
+        }
+    }
+
+    /// Evaluate the current state and make transitions as needed; returns false if we should shut down
+    fn state_machine_tick(&mut self) -> bool {
+        // Increment the state machine using latest gathered data
+        let prev_state = self.current_tpu_state;
+        self.current_tpu_state = self.get_next_tpu_state();
+
+        datapoint_info!(
```
haoranjito: This datapoint is emitted every slot ms. could be too much?
Maybe we emit only when state changed?

core/src/proxy/fetch_stage_manager.rs
```diff
@@ -82,173 +44,702 @@ impl FetchStageManager {
...
+                prev_state, self.current_tpu_state
+            );
+            self.gossip_current_tpu_state()
+        } else {
+            true
+        }
+    }
+
+    /// Log metrics and reset counters; returns false if we should shut down
+    fn handle_metrics_tick(&mut self) -> bool {
```
haoranjito: nit: this fn always return true. Can we remove the return value from this fn?

core/src/proxy/fetch_stage_manager.rs
```diff
@@ -82,173 +44,702 @@ impl FetchStageManager {
...
+            self.relayer_info = None;
+        }
+
+        // Update gossip if the state changed
+        if prev_state != self.current_tpu_state {
+            info!(
+                "Switching TPU state from {:?} to {:?}",
+                prev_state, self.current_tpu_state
+            );
+            self.gossip_current_tpu_state()
```
haoranjito: When gossip_current_tpu_state fails, the new code will break the loop and exit.
while for the original code, it just log an error and continue. 

Maybe it is okay. We expect that the socket address are always well formed?

core/src/proxy/fetch_stage_manager.rs
```diff
@@ -82,173 +44,702 @@ impl FetchStageManager {
...
+        assert_eq!(brain.current_tpu_state.fwd_addr, *expected_fwd_addr);
+    }
+
+    fn check_cluster_info(
+        cluster_info: &ClusterInfo,
+        expected_addr: &SocketAddr,
+        expected_fwd_addr: &SocketAddr,
+    ) {
+        assert_eq!(
+            cluster_info.my_contact_info().tpu(Protocol::QUIC).unwrap(),
```
haoranjito, line 482: Note that we should use UDP here. 

```suggestion
            cluster_info.my_contact_info().tpu(Protocol::UDP).unwrap(),
```

cluster_info is populated with the UDP port, so we should be using UDP here.

Note: This test is currently passing on master due to a bug where cluster_info incorrectly uses QUIC for the UDP port. The two bugs effectively cancel out, so the test appears to pass. Once we sync with agave-master (i.e. that bug is fixed), this test will fail and we will need to correct it then. 

We don't need to do anything now. Just mark it for future fix.

core/src/proxy/fetch_stage_manager.rs
```diff
@@ -82,173 +44,702 @@ impl FetchStageManager {
...
+            *expected_addr
+        );
+        assert_eq!(
+            cluster_info
+                .my_contact_info()
+                .tpu_forwards(Protocol::QUIC)
```
haoranjito, line 488: Don't need to commit now. For future fix.

```suggestion
                .tpu_forwards(Protocol::UDP)
```

core/src/proxy/fetch_stage_manager.rs
```diff
@@ -82,173 +44,702 @@ impl FetchStageManager {
...
+    }
+
+    fn setup_test() -> TestContext {
+        let cluster_info = new_test_cluster_info();
+        let bam_enabled = Arc::new(AtomicU8::new(BamConnectionState::Disconnected as u8));
+        let original_tpu_info = TpuAddresses {
+            tpu_addr: cluster_info.my_contact_info().tpu(Protocol::QUIC).unwrap(),
+            tpu_forward_addr: cluster_info
+                .my_contact_info()
+                .tpu_forwards(Protocol::QUIC)
```
haoranjito, line 528: Don't need to commit now. For future fix.

```suggestion
                .tpu_forwards(Protocol::UDP)
```

core/src/proxy/fetch_stage_manager.rs
```diff
@@ -82,173 +44,702 @@ impl FetchStageManager {
...
+        _packet_tx: Sender<PacketBatch>,
+        packet_rx: Receiver<PacketBatch>,
+        brain: FetchStageTpuStateMachine,
+    }
+
+    fn setup_test() -> TestContext {
+        let cluster_info = new_test_cluster_info();
+        let bam_enabled = Arc::new(AtomicU8::new(BamConnectionState::Disconnected as u8));
+        let original_tpu_info = TpuAddresses {
+            tpu_addr: cluster_info.my_contact_info().tpu(Protocol::QUIC).unwrap(),
```
haoranjito, line 525: Don't need to commit now. For future fix.

```suggestion
            tpu_addr: cluster_info.my_contact_info().tpu(Protocol::UDP).unwrap(),
```

core/src/proxy/fetch_stage_manager.rs
segfaultdoc: updated to revert state and continue

PR #1180, opened 2026-02-10
ottersec: Relax insert_new_batch to fill up the slab fully

haoranjito:
#### Problem

We can fill up the slab fully when insert_new_batch. insert_new_batch is
only called in bam scheduler. And bam_scheduler doesn't call insert
single transaction (i.e. insert single transaction may panic if the slab
is full). Therefore, it is safe to relax insert_new_batch fo fill up the
slab.

#### Summary of Changes

Fixes #


haoranjito:
Fixed in https://github.com/jito-foundation/jito-solana/pull/1215

PR #1181, opened 2026-02-10
ottersec: Fix reporting for multi‑tx non‑revert bundles

haoranjito:
#### Problem

Reporting for non revert batches was incorrect: it always returned the
first transaction's result and ignored the rest. This change updates
reporting so non‑revert multi‑tx batches return a single Committed
result containing all transaction results when the batch fully commits.
If any transaction fails, the response stays NotCommitted with the first
failure index (the only shape the AtomicTxnBatchResult proto can
represent for mixed outcomes).

This is an intentional tradeoff given the proto limitation: it cannot
encode partial failures alongside committed results.

#### Summary of Changes

Fixes #


haoranjito:
This is not an issue. On the server side, only single tx batch is allowed when no-revert = true.

PR #1207, opened 2026-02-12
Backport fix fetch stage manager

segfaultdoc:
Backports #1172 

This was tested on london-mainnet-validator-1 (Jito1). Toggled between bam disabled and relayer enabled successfully with bam-enabled always taking priority.

### DIFF
--- a/core/src/proxy/fetch_stage_manager.rs
+++ b/core/src/proxy/fetch_stage_manager.rs
@@ -22,7 +22,7 @@ use {
     },
 };
 
-/// How often to check for heartbeat timeouts
+/// Maximum time allowed between relayer heartbeats before considering the relayer timed out
 const MAX_TIME_BETWEEN_RELAYER_HEARTBEATS: Duration = Duration::from_millis(1500); // Empirically determined from load testing
 
 /// How often to re-evaluate TPU state
@@ -310,13 +310,6 @@ impl FetchStageTpuStateMachine {
         let state_changed = prev_state != self.current_tpu_state;
         let report_due = Instant::now().duration_since(self.last_tpu_report) >= TPU_REPORT_INTERVAL;
 
-        // Reset relayer info if we switched away from it;
-        if prev_state.tpu_type == TpuConnectionType::Relayer
-            && self.current_tpu_state.tpu_type != TpuConnectionType::Relayer
-        {
-            self.relayer_info = None;
-        }
-
         let mut succeeded = true;
 
         // Update gossip if the state changed
@@ -332,6 +325,12 @@ impl FetchStageTpuStateMachine {
                 self.current_tpu_state = prev_state;
                 let _ = self.gossip_current_tpu_state();
                 succeeded = false;
+            }
+            if succeeded
+                && prev_state.tpu_type == TpuConnectionType::Relayer
+                && self.current_tpu_state.tpu_type != TpuConnectionType::Relayer
+            {
+                self.relayer_info = None;
             }
         }
 
@@ -351,7 +350,7 @@ impl FetchStageTpuStateMachine {
         succeeded
     }
 
-    /// Log metrics and reset counters; returns false if we should shut down
+    /// Log metrics and reset counters
     fn handle_metrics_tick(&mut self) {
         datapoint_info!(
             "relayer-heartbeat",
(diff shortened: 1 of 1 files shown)


core/src/proxy/fetch_stage_manager.rs
```diff
@@ -82,173 +44,702 @@ impl FetchStageManager {
...
+        assert_eq!(brain.current_tpu_state.fwd_addr, *expected_fwd_addr);
+    }
+
+    fn check_cluster_info(
+        cluster_info: &ClusterInfo,
+        expected_addr: &SocketAddr,
+        expected_fwd_addr: &SocketAddr,
+    ) {
+        assert_eq!(
+            cluster_info.my_contact_info().tpu(Protocol::QUIC).unwrap(),
```
haoranjito:
```suggestion
            cluster_info.my_contact_info().tpu(Protocol::UDP).unwrap(),
```

This should fix the test. cluster_info is populated with the UDP port, so we should be using UDP here.

Note: This test is currently passing on master due to a bug where cluster_info incorrectly uses QUIC for the UDP port. The two bugs effectively cancel out, so the test appears to pass. Once we sync with agave-master (i.e. that bug is fixed), this test will fail and we will need to correct it then.

core/src/proxy/fetch_stage_manager.rs
```diff
@@ -82,173 +44,702 @@ impl FetchStageManager {
...
+            *expected_addr
+        );
+        assert_eq!(
+            cluster_info
+                .my_contact_info()
+                .tpu_forwards(Protocol::QUIC)
```
haoranjito:
```suggestion
                .tpu_forwards(Protocol::UDP)
```

core/src/proxy/fetch_stage_manager.rs
```diff
@@ -82,173 +44,702 @@ impl FetchStageManager {
...
+        _packet_tx: Sender<PacketBatch>,
+        packet_rx: Receiver<PacketBatch>,
+        brain: FetchStageTpuStateMachine,
+    }
+
+    fn setup_test() -> TestContext {
+        let cluster_info = new_test_cluster_info();
+        let bam_enabled = Arc::new(AtomicU8::new(BamConnectionState::Disconnected as u8));
+        let original_tpu_info = TpuAddresses {
+            tpu_addr: cluster_info.my_contact_info().tpu(Protocol::QUIC).unwrap(),
```
haoranjito:
```suggestion
            tpu_addr: cluster_info.my_contact_info().tpu(Protocol::UDP).unwrap(),
```

core/src/proxy/fetch_stage_manager.rs
```diff
@@ -82,173 +44,702 @@ impl FetchStageManager {
...
+            tpu_forward_addr: cluster_info
+                .my_contact_info()
+                .tpu_forwards(Protocol::QUIC)
```
haoranjito:
```suggestion
                .tpu_forwards(Protocol::UDP)
```

haoranjito:
looks like we may need to fix the type after the rebase
```

error[E0308]: mismatched types
--
--> core/tests/bam_connection.rs:540:13
\|
538 \|         let _manager = BamManager::new(
\|                        --------------- arguments to this function are incorrect
539 \|             exit.clone(),
540 \|             bam_url,
\|             ^^^^^^^ expected `Arc<ArcSwapAny<Arc<Option<String>>>>`, found `Arc<Mutex<Option<String>>>`
\|
= note: expected struct `std::sync::Arc<arc_swap::ArcSwapAny<std::sync::Arc<std::option::Option<std::string::String>>>>`
found struct `std::sync::Arc<std::sync::Mutex<std::option::Option<std::string::String>>>`
note: associated function defined here
--> /solana/core/src/bam_manager.rs:71:12


```

core/src/proxy/fetch_stage_manager.rs
```diff
@@ -82,173 +45,760 @@ impl FetchStageManager {
...
+
+        let state_changed = prev_state != self.current_tpu_state;
+        let report_due = Instant::now().duration_since(self.last_tpu_report) >= TPU_REPORT_INTERVAL;
+
+        // Reset relayer info if we switched away from it;
+        if prev_state.tpu_type == TpuConnectionType::Relayer
+            && self.current_tpu_state.tpu_type != TpuConnectionType::Relayer
+        {
+            self.relayer_info = None;
+        }
```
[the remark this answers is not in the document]
haoranjito: @segfaultdoc yeah. I think this is a corner case that we may need to fix.

HaoranYi (approved):
:shipit:

PR #1211, opened 2026-02-12
Only check the fee payer on transaction index 0 for bundles

buffalu:
#### Problem
This is a regression from v3.0 and how the bundle execution works. Some bundles will seed fee payers in transactions 1..N. 

#### Summary of Changes
Only check the fee payer on the first transaction


buffalu:
> lgtm.
> 
> I have the same fix in #1236. I also added a test for this in the above PR.

ah ok, lets take yours!

PR #1213, opened 2026-02-17
Clean up bam connection handling

esemeniuc:
Goal: make bam connection thrash less, and easier to follow

 - Migrated shared BAM URL state from Mutex<Option<String>> to ArcSwap<Option<String>> in admin svc. This makes BAM URL reads/writes lock-free in hot paths and RPC handlers, and avoids mutex contention during concurrent updates.
  - Refactored BamManager::run (core/src/bam_manager.rs) into disconnected and connected flows with a central handle_identity_change path; this makes reconnect sequencing deterministic and easier to reason about under concurrent identity/url changes.
  - Fixed identity-change reconnect ordering bug in core/src/bam_manager.rs: on identity update, the manager now disconnects immediately, then waits for cluster_info identity propagation only while disconnected; this prevents holding a live BAM connection under the old identity while waiting, which could lead to stale-identity behavior during auth/reconnect
  - Fixed identity-change synchronization race by switching from plain relaxed flag checks to swap(false, Ordering::AcqRel) and Release stores (core/src/bam_manager.rs); this was done to prevent missed or delayed reconnect signals across threads.
  - Updated identity wait logic to follow the latest requested identity while waiting in bam_manager.rs::wait_for_identity_in_cluster_info; this fixes stale-target waits when identities change multiple times quickly, which could previously wait for an outdated key and timeout unnecessarily.
  - Made blocking waits exit-aware in both core/src/bam_connection.rs and core/src/bam_manager.rs; this speed up shutdown, rather than wait until timeout (health/config wait and identity-propagation wait)
  - Clarified BAM state transitions (Disconnected/Connecting/Connected) and URL-change disconnect behavior in the new loop structure (core/src/bam_manager.rs); this keeps downstream state consumers consistent and reduce ambiguous “connection lost” handling.
  - Add connection timeouts

### DIFF
--- a/core/src/bam_manager.rs
+++ b/core/src/bam_manager.rs
@@ -96,7 +96,7 @@ impl BamManager {
         identity_notifiers: Arc<RwLock<KeyUpdaters>>,
     ) {
         let runtime = tokio::runtime::Builder::new_multi_thread()
-            .worker_threads(2)
+            .worker_threads(8)
             .enable_all()
             .build()
             .unwrap();
@@ -433,8 +433,8 @@ impl BamManager {
     }
 
     fn wait_for_identity_in_cluster_info(
-        new_identity: &Arc<ArcSwap<Option<Pubkey>>>,
-        cluster_info: &Arc<ClusterInfo>,
+        new_identity: &ArcSwap<Option<Pubkey>>,
+        cluster_info: &ClusterInfo,
         exit: &AtomicBool,
         timeout: std::time::Duration,
     ) -> bool {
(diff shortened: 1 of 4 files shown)


core/src/bam_manager.rs
```diff
@@ -395,33 +433,47 @@ impl BamManager {
     }
 
     fn wait_for_identity_in_cluster_info(
-        new_identity: Option<Pubkey>,
+        new_identity: &Arc<ArcSwap<Option<Pubkey>>>,
```
haoranjito: nit: &Arc is anti-pattern. 

since this is not a hot-loop, i think passing Arc will be fine. 
```suggestion
        new_identity: Arc<ArcSwap<Option<Pubkey>>>,
        cluster_info: Arc<ClusterInfo>,
```

core/src/bam_manager.rs
```diff
@@ -96,13 +90,13 @@ impl BamManager {
...
         exit: Arc<AtomicBool>,
-        bam_url: Arc<Mutex<Option<String>>>,
+        bam_url: Arc<ArcSwap<Option<String>>>,
         dependencies: BamDependencies,
         poh_recorder: Arc<RwLock<PohRecorder>>,
         identity_notifiers: Arc<RwLock<KeyUpdaters>>,
     ) {
         let runtime = tokio::runtime::Builder::new_multi_thread()
-            .worker_threads(8)
+            .worker_threads(2)
```
haoranjito: Why do we reduce the worker threads to 2? 
Are 2 threads enough to handle all the async works in the connection tasks (i.e. health checks, heartbeats, gRPC)?

esemeniuc: reverted

core/src/bam_manager.rs
esemeniuc: fixed

PR #1215, opened 2026-02-17
Fix insert_new_batch slab-capacity check

esemeniuc:
#### Problem
  - In core/src/banking_stage/transaction_scheduler/transaction_state_container.rs, it used:
      - capacity_required = len + txns_len + 1
      - if capacity_required >= capacity { return None; }

  That rejects exact-fit inserts (capacity_required == capacity), even though they are valid. So batches could be incorrectly dropped despite enough space.

[Slab docs:](https://docs.rs/slab/latest/slab/#:~:text=If%20a%20slab%E2%80%99s%20length%20is%20equal%20to%20its%20capacity%2C%20the%20next%20value%20inserted%20into%20the%20slab%20will%20require%20growing%20the%20slab%20by%20reallocating)
```
If a slab’s length is equal to its capacity, the next value inserted into the slab will require growing the slab by reallocating.
```

#### Summary of Changes
  - Compute:
      - entries_required = txns_max_age.len() + 1 (batch metadata + tx entries)
      - available_entries = capacity - len
  - Reject only when entries_required > available_entries.
  - Use saturating arithmetic for defensive correctness.

Credit to @haoranjito for finding the bug

Fixes #

### DIFF
--- a/core/src/banking_stage/transaction_scheduler/transaction_state_container.rs
+++ b/core/src/banking_stage/transaction_scheduler/transaction_state_container.rs
@@ -89,7 +89,7 @@ pub(crate) trait StateContainer<Tx: TransactionWithMeta> {
     /// Panics if the transaction does not exist.
     fn get_transaction(&self, id: TransactionId) -> Option<&Tx>;
 
-    /// Get the batch id and revert_on_error flag for a transaction.
+    /// Get the transaction ids and metadata for a batch id.
     ///
     /// `BamReceiveAndBuffer::prevalidate_batches` rejects `AtomicTxnBatch`es with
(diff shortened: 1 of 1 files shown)


core/src/banking_stage/transaction_scheduler/transaction_state_container.rs
```diff
@@ -717,4 +720,60 @@ mod tests {
...
+                revert_on_error: false,
+                max_schedule_slot: 0,
+            }));
+            container
+                .batch_id_to_transaction_ids
+                .insert(batch_id, SmallVec::new());
+        }
+    }
+
+    #[test]
```
haoranjito, line 742: 👍 for adding these tests.

PR #1244, opened 2026-02-23
Proxy stage metric fix

segfaultdoc:
#### Problem
Printing the entire Status struct to influx breaks all metrics being submitted in the same batch. When the error is emitted frequently then effectively no metrics show up.

#### Summary of Changes
Simplify the error being printed.

### DIFF
--- a/core/src/proxy/mod.rs
+++ b/core/src/proxy/mod.rs
@@ -128,7 +128,7 @@ mod tests {
     #[test]
     fn test_sanitize_status_message_for_influx() {
         let input = r#"Relayer is unhealthy!!! @@@ Please "try" \again\ when {it's} [healthy]."#;
-        let expected = "Relayer is unhealthy Please try again when its healthy";
+        let expected = "Relayer is unhealthy  Please try again when its healthy";
         assert_eq!(sanitize_status_message_for_influx(input), expected);
     }
 }


core/src/proxy/mod.rs
```diff
@@ -29,8 +29,8 @@ type HeartbeatEvent = (SocketAddr, SocketAddr);
 
 #[derive(Error, Debug)]
 pub enum ProxyError {
-    #[error("grpc error: {0}")]
-    GrpcError(Box<Status>),
+    #[error("GrpcError: code={code}, message={message}")]
+    GrpcError { code: tonic::Code, message: String },
```
haoranjito, line 33: maybe sanitize the message too? 

```
pub fn sanitize_message(message: &str) -> String {
      message
          .replace('\\', "\\\\")
          .replace('\n', "\\n")
          .replace('\r', "\\r")
          .replace('\t', "\\t")
  }

ProxyError::GrpcError {
      code: status.code(),
      message: sanitize_message(status.message()),
  }
```

ebin-mathews: Do we need something like

```
/// To avoid errors in https://github.com/jito-foundation/jito-solana/blob/fd776ab77a50bc4699c7b46c2c7b295fa8aef984/metrics/src/metrics.rs#L164C28-L164C56
pub fn solana_metrics_sanitize_str(input: &str, limit: usize) -> String {
    input
        .chars()
        .filter(|c| matches!(c, 'a'..='z' | 'A'..='Z' | '0'..='9' | '_' | '-'))
        .take(limit)
        .collect()
}
```

We do this in block engine

haoranjito (approved):
:shipit:

PR #1274, opened 2026-03-02
Fix race for secondary index (v3.1)

haoranjito:
#### Problem

- **Patches race in secondary index w.r.t. insert and remove**
 
Manually ported https://github.com/anza-xyz/agave/pull/10964 to v3.1 jito-solana since the anza upstream PR hasn't back ported to v3.1 yet. 

This patch fix a race condition when secondary index is enabled on the validator. 

#### Summary of Changes

Fixes #


0xEdgar (approved):
do we need this for v4.0 as well?

haoranjito:
> do we need this for v4.0 as well?

No. it has been backported to v4.0 in anza.

0xEdgar:
going to close given we are moving to v4.0 and it is present there https://github.com/anza-xyz/agave/pull/10965

PR #1335, opened 2026-03-16
runtime: return logs and compute units in simulate_bundle on failure

SAMIR897:
#### Summary of Changes

Fixes #


SAMIR897 (commented):
solution

PR #1411, opened 2026-04-10
turbine: route multicast shreds through unbound socket

mindrunner:
When --bind-address is set, broadcast and retransmit sockets are bound to that IP. The kernel's route lookup then uses the bound source IP as a constraint, forcing multicast traffic out the bound interface regardless of the multicast routing table.

Fix: in broadcast_shreds() and retransmit_shred(), partition packets by is_multicast() and send multicast destinations through a thread-local UDP socket bound to 0.0.0.0:0 (with TTL=64). The kernel then picks the correct egress interface from its routing table. Unicast traffic continues to use the original bound sockets.


esemeniuc:
i think this is covered in https://github.com/jito-foundation/jito-solana/pull/1419 (with less allocs)

PR #1412, opened 2026-04-10
turbine: use a dedicated 0.0.0.0-bound socket for ShredReceiverAddresses

juan-malbeclabs:
# Problem

Turbine's broadcast sockets are created bound to the IP address specified by `--bind-address` (the primary interface). When the validator forwards shreds to destinations configured via `--shred-receiver-address` and `multicast_receiver_address`, it reuses that same interface-locked socket.

This causes a silent routing failure: if any of those destinations are reachable through a different network interface than the one specified by `--bind-address`, the OS will either drop the packets at the NIC driver level or send them out the wrong interface, and they will never reach the destination.

When XDP is enabled (`--experimental-retransmit-xdp-cpu-cores`), the problem is worse: XDP bypasses the kernel network stack entirely and is attached to the specific NIC corresponding to `--bind-address`. It has no visibility into the routing table, so external receiver packets would be silently dropped.

# Solution

Create a dedicated `UdpSocket` bound to `0.0.0.0` exclusively for forwarding shreds to `ShredReceiverAddresses` and `multicast_receiver_address`. Because the socket is not bound to any specific interface, the kernel IP stack uses the routing table to select the correct outbound interface for each destination.

This socket is created once in `BroadcastStage::new()`, wrapped in `Arc` and shared across all transmit threads. It uses `batch_send` regardless of whether the main broadcast path is UDP or XDP.

Turbine tree traffic and shredstream forwarding are unaffected, they continue using the main socket as before.

# Changes

- `BroadcastStage::new()`: creates one `Arc<UdpSocket>` bound to `0.0.0.0` with multicast TTL 64, shared across all transmit threads.
- `broadcast_shreds()`: collects external receiver packets (`ShredReceiverAddresses` + multicast) into a separate `external_packets  vec and sends them via the dedicated socket using  `batch_send`. This applies to both the UDP and XDP code paths.
- `BroadcastRun::transmit()` trait and all implementors: updated to accept and forward the new `shred_receiver_socket` parameter.


haoranjito:
v4.0 is updated. 
we may need to rebase - skip the jito patch and v4.0 commit.

PR #1424, opened 2026-04-20
[MEV-8694] feat: publish near leader shreds to bampc

haoranjito:
#### Problem

Publish near leader shreds to bampc

#### Summary of Changes
Add feature to publish near leader shreds to bampc.

- propagate bamconfig.shred_socks
- tpu broadcast leader shreds to bam endpoints
- tvu retransmit near leader shreds to bam endpoints (2 slots before its leader slot).

#### Test
1. `bam_shred` config propagated to validator 

```
[2026-04-21T00:37:50.320390000Z INFO  solana_core::bam_manager] Setting 1 BAM shred receiver address(es) from BAM config
```

2. jito validator forwarded shreds to BAM

leader broadcast
```
[2026-04-21T00:38:00.228487000Z INFO  solana_turbine::broadcast_stage] broadcast_shreds: slots=[37] bam_shred_receiver_addresses=[127.0.0.1:12001] additional_external_addrs=[127.0.0.1:12001]
```

pre-leader retransmit 
```
[2026-04-21T00:37:50.740889000Z INFO  solana_turbine::retransmit_stage] retransmit batch: batch_slots=[18] bam_forward_slots=[18] bam_shred_receiver_addresses=[127.0.0.1:12001]
  [2026-04-21T00:37:50.740916000Z INFO  solana_turbine::retransmit_stage] retransmit_shred: slot=18 index=0 including bam_shred_receiver_addresses=[127.0.0.1:12001]
```

3. bam received forwarded shreds

```
[2026-04-21T00:38:12.286664Z INFO  solana_metrics::metrics] datapoint: shred_fetch index_overrun=0i shred_count=2433i ...
  [2026-04-21T00:38:12.286694Z INFO  solana_metrics::metrics] datapoint: validator_service_metrics heartbeats=0i auth_proofs=0i auth_challenges=0i leader_states=33i ...

```

Fixes #

### DIFF
--- a/core/src/bam_manager.rs
+++ b/core/src/bam_manager.rs
@@ -32,6 +32,7 @@ use {
     solana_runtime::bank::Bank,
     solana_signer::Signer,
     solana_tls_utils::NotifyKeyUpdate,
+    solana_turbine::{MAX_SHRED_RECEIVER_ADDRESSES, ShredReceiverAddresses},
     solana_version::ClientId,
 };
 
@@ -155,6 +156,7 @@ impl BamManager {
                         dependencies
                             .bam_enabled
                             .store(BamConnectionState::Disconnected as u8, Ordering::Relaxed);
+                        Self::clear_bam_shred_receiver_addresses(&dependencies);
                         std::thread::sleep(WAIT_TO_RECONNECT_DURATION);
                         continue;
                     };
@@ -175,6 +177,7 @@ impl BamManager {
                             dependencies
                                 .bam_enabled
                                 .store(BamConnectionState::Disconnected as u8, Ordering::Relaxed);
+                            Self::clear_bam_shred_receiver_addresses(&dependencies);
                             std::thread::sleep(WAIT_TO_RECONNECT_DURATION);
                             continue;
                         }
@@ -194,6 +197,7 @@ impl BamManager {
                         dependencies
                             .bam_enabled
                             .store(BamConnectionState::Disconnected as u8, Ordering::Relaxed);
+                        Self::clear_bam_shred_receiver_addresses(&dependencies);
                         std::thread::sleep(WAIT_TO_RECONNECT_DURATION);
                         continue;
                     }
@@ -201,6 +205,7 @@ impl BamManager {
                     info!("BAM connection established");
                     if let Some(builder_config) = connection.get_latest_config() {
                         Self::update_tpu_config(Some(&builder_config), &dependencies);
+                        Self::update_shred_socks_config(Some(&builder_config), &dependencies);
                         Self::update_block_engine_key_and_commission(
                             Some(&builder_config),
                             &dependencies.block_builder_fee_info,
@@ -221,6 +226,7 @@ impl BamManager {
                 dependencies
                     .bam_enabled
                     .store(BamConnectionState::Disconnected as u8, Ordering::Relaxed);
+                Self::clear_bam_shred_receiver_addresses(&dependencies);
                 warn!("BAM connection unhealthy");
                 continue;
             }
@@ -244,6 +250,7 @@ impl BamManager {
                 dependencies
                     .bam_enabled
                     .store(BamConnectionState::Disconnected as u8, Ordering::Relaxed);
+                Self::clear_bam_shred_receiver_addresses(&dependencies);
                 if let Some(new_url) = configured_bam_url.as_deref() {
                     info!("BAM URL changed, connecting to new URL: {new_url}");
                 } else {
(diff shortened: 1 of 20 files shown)


buffalu:
can we piggy back to using the current shred receivers + retransmit instead?

haoranjito:
> can we piggy back to using the current shred receivers + retransmit instead?

maybe not. bam shred endpoints are owned by bampc config and updated by bam connection manager. existing shred receiver set has different ownership/lifecycle. and different send policy for bam shreds, i.e. they are only forwarded near the validator's leader window.

core/src/bam_manager.rs
```diff
@@ -359,9 +368,10 @@ impl BamManager {
     fn get_sockaddr(info: Option<&Socket>) -> Option<SocketAddr> {
         let info = info?;
         let Socket { ip, port } = info;
+        let port = u16::try_from(*port).ok().filter(|port| *port != 0)?;
```
0xEdgar: why do we need this?

turbine/src/retransmit_stage.rs
```diff
@@ -510,10 +575,20 @@ fn retransmit_shred(
...
+                    .saturating_add(shred_receiver_addresses.len())
+                    .saturating_add(if include_bam_shred_receivers {
+                        bam_shred_receiver_addresses.len()
+                    } else {
+                        0
+                    }),
+            );
             send_addrs.extend(addrs.iter().copied());
             send_addrs.extend(shred_receiver_addresses.iter().copied());
+            if include_bam_shred_receivers {
```
0xEdgar: so this extends the max shred recv addresses from 32-> 64, any performance implications there?

turbine/src/retransmit_stage.rs
```diff
@@ -488,13 +542,24 @@ fn retransmit_shred(
...
+                send_addrs.extend(bam_shred_receiver_addresses.iter().copied());
+            }
             if !send_addrs.is_empty()
```
0xEdgar: is there dedup being done between all these send addrs?

0xEdgar: if so i think we should dedupe, maybe in retransmit + broadcast

turbine/src/retransmit_stage.rs
```diff
@@ -456,6 +468,46 @@ fn retransmit(
...
+            .map(|leader| leader.id)
+    })
+}
+
+fn should_forward_to_bam_for_slot_with_leader_lookup(
+    slot: Slot,
+    my_pubkey: &Pubkey,
+    mut leader_at_slot: impl FnMut(Slot) -> Option<Pubkey>,
+) -> bool {
+    (0..3).any(|offset| {
```
0xEdgar, line 491: might be some weirdness with 2 slots before epoch boundary here, but probably fine

core/src/bam_manager.rs
haoranjito: more like defensive protection in case bampc config is wrong. 
if we trust bampc config, we don't need it. 
reverted.

turbine/src/retransmit_stage.rs
haoranjito: yes. that's worst case. 
practically, bam_shred_receiver won't be 32 though. most likely, a few.

turbine/src/retransmit_stage.rs
haoranjito: yeah. broadcast already did dup. 
i will add dup here too.

turbine/src/retransmit_stage.rs
haoranjito: agreed. probably fine. 

I think maybe it worth to increase it to 4 to cover shreds from all previous leader? but fire-bam is using 2. so we choose 2 here.

buffalu:
why not add bam to shred receiver address and dynamically add/remove from the retransmit addresses close to leader slot? less plumbing changes and keep all the logic in bam manager

haoranjito:
> why not add bam to shred receiver address and dynamically add/remove from the retransmit addresses close to leader slot? less plumbing changes and keep all the logic in bam manager

It seems this would put more responsibility into bam_manager and require exporting more bank/slot state into that path. bam_manager would need to know leader schedule and slot context in order to dynamically add/remove BAM shred receivers around the leader window. It feels like more work than patching the TVU/TPU send paths directly where the slot/leader context already exists.

esemeniuc (approved):

0xEdgar (approved):
lgtm

haoranjito:
merged with #1431

PR #1427, opened 2026-04-23
[Master] Poh Core ordering

esemeniuc:
Issue: POH core can conflict with other tasks pinned to core 0 on host


haoranjito (approved):
lgtm.

PR #1431, opened 2026-04-27
[Master] Shred rework

esemeniuc:
## Summary
- Forward BAM-provided shred receiver sockets through broadcast and retransmit paths.
- Gate BAM retransmit forwarding to slots just before this validator’s leader window, with de-duping and receiver caps.
- Wire BAM shred receiver state through TPU/TVU and clear it on BAM disconnect.
- Add focused tests for external receiver de-duping, caps, and BAM gating.

### DIFF
--- a/turbine/src/broadcast_stage.rs
+++ b/turbine/src/broadcast_stage.rs
@@ -33,7 +33,7 @@ use {
     solana_streamer::sendmmsg::{SendPktsError, batch_send},
     solana_time_utils::{AtomicInterval, timestamp},
     std::{
-        collections::{HashMap, HashSet},
+        collections::HashSet,
         net::{SocketAddr, UdpSocket},
         sync::{
             Arc, Mutex, RwLock,
@@ -129,6 +129,7 @@ impl BroadcastStageType {
         votor_event_sender: VotorEventSender,
         shredstream_receiver_address: Arc<ArcSwap<Option<SocketAddr>>>,
         shred_receiver_addresses: Arc<ArcSwap<ShredReceiverAddresses>>,
+        bam_shred_receiver_addresses: Arc<ArcSwap<ShredReceiverAddresses>>,
         multicast_receiver_address: Arc<ArcSwap<Option<SocketAddr>>>,
     ) -> BroadcastStage {
         let migration_status = bank_forks.read().unwrap().migration_status();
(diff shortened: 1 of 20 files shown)


turbine/src/broadcast_stage.rs
```diff
@@ -640,30 +682,10 @@ pub fn broadcast_shreds(
             }
             send_xdp_time.stop();
             transmit_stats.send_xdp_elapsed += send_xdp_time.as_us();
```
haoranjito, line 674: what about external_packets when xdp is enabled? are we not sending them?

turbine/src/broadcast_stage.rs
```diff
@@ -552,68 +567,87 @@ pub fn broadcast_shreds(
...
+
+    for &addr in external_addrs
         .iter()
-        .filter(|addr| shredstream_receiver_address.as_ref() != Some(addr))
-        .chain(multicast_receiver_address.iter().filter(|addr| {
-            shred_receiver_addresses.len() < MAX_SHRED_RECEIVER_ADDRESSES
-                && !shred_receiver_addresses.contains(addr)
-                && shredstream_receiver_address.as_ref() != Some(addr)
-        }))
+        .filter(|addr| Some(*addr) == shredstream_receiver_address)
```
haoranjito: is shredstream_receiver_address special? seems that it is included in the main_packet? while others are treated as external packet?

turbine/src/retransmit_stage.rs
```diff
@@ -487,19 +533,49 @@ fn retransmit_shred(
...
     let num_nodes = match socket {
         RetransmitSocket::Xdp(sender) => {
             let mut sent = num_addrs;
             let mut send_addrs =
-                Vec::with_capacity(num_addrs.saturating_add(shred_receiver_addresses.len()));
-            send_addrs.extend(addrs.iter().copied());
-            send_addrs.extend(shred_receiver_addresses.iter().copied());
+                Vec::with_capacity(turbine_addrs.len().saturating_add(external_addrs.len()));
+            send_addrs.extend(turbine_addrs.iter().copied());
+            send_addrs.extend(external_addrs.iter().copied());
```
haoranjito, line 573: Seems that we are now sending external_addrs with xdp? 
I think it should work. but we need to test in xdp to make sure that bam and other shred endpoint can be reached by xdp.

turbine/src/broadcast_stage.rs
haoranjito: may want to add shredstream_receiver_address and bam_shred_reciever address to xdp send? 
I think they should be reachable by xdp. we need to test that under xdp.

haoranjito (approved):
lgtm.

PR #1438, opened 2026-05-02
[master] Port EntryBytesBudget to normal and bundle consumers

### LINKED ISSUE #1421: Merge the entry_bytes changes into master for both normal consumer and bundle consumer
#### Problem
v3.0 and v4.0 recently fixed a bug related to adding a cap to the number of bytes in all entries. This change was only merged into v3.0 and v4.0 and needs to be merged into mater

shubhamessier:
#### Problem                                                                                                                                                                                                     
v3.0 and v4.0 received a fix that caps the total entry bytes recorded per slot, but it was never merged into master. Without it, leaders on master can pack arbitrarily many bytes into a slot, producing  oversized blocks that consensus/replay rejects.                   
                                                                                                                                                                                                                   
#### Summary of Changes                                        
- Add `EntryBytesBudget` on `Bank` that caps total entry bytes recorded per slot at 20 MiB; child banks inherit the slot limit from the parent.
- Wire the budget into the banking-stage consumer (`record_batch`) and the bundle consumer (`record_bundle`); over-budget batches map `EntryBytesReserveError::ExceedsSlotLimit` to
  `PohRecorderError::MaxHeightReached` so the affected transactions retry through the existing recorder-error path.
- Add `TransactionWithMeta::serialized_size()` so consumers compute the byte cost without re-serializing.
- Ports upstream agave #11865 (v3.1 backport 437252fc43) plus the bundle-side commit 00daff2f39, adapted to master's `record_batch` recorder which produces N entries per call.
                                                           
Fixes #1421


shubhamessier:
Hey @apfitzge can you kindly review the changes, Thanks!

cc: @buffalu

core/src/bundle_stage/bundle_consumer.rs
```diff
@@ -411,12 +419,15 @@ impl BundleConsumer {
             attempted_processing_count: processing_results.len() as u64,
         };
 
+        let mut entry_bytes =
```
apfitzge, line 422: has master not moved to using a single entry for bundles yet? If it has, we do not need to do the multiplication here.

apfitzge (approved):

core/src/bundle_stage/bundle_consumer.rs
shubhamessier: Not yet actually, on master record_bundle in poh/src/transaction_recorder.rs:163 still pushes one vec![tx] per transaction, and PohRecorder::record (poh/src/poh_recorder.rs:357) emits one Entry per batch, so an N-tx bundle still produces N entries. I'll drop the multiplication the moment the single-entry-per-bundle change lands.

shubhamessier: Hey @apfitzge, let me know if you'd like any changes. Thanks!

apfitzge: that's fine. It's not really related to this change.

Ping someone from Jito for an actual review - I cannot give approval to merge here.

shubhamessier: Sure

shubhamessier: @buffalu can you have a look at the changes once, thanks!

shubhamessier: Hey @romeo4934 @steveluscher @jayvdb the pr as been stale, can you have a look at the changes, thanks!

esemeniuc:
Fixed in https://github.com/jito-foundation/jito-solana/pull/1536

PR #1441, opened 2026-05-06
Refactor build pipeline to drive upstream tooling under BuildKit

jito-spaceman:
#### Problem

Upstream tooling has changed significantly underneath the Jito-developed `dev/Dockerfile` + `f` build scripts making the latter no longer relevant / easily maintainable, as well as implementation of the latter needing modernization/cleanup since originally created.

#### Summary of Changes

Replace `dev/Dockerfile` + `f` with a BuildKit multi-stage build that drives `scripts/create-release-tarball.sh` + `cargo-install-all.sh` and exports a Jito-curated operator artifact to `./dist/`. Adds `f --checkout REF` for building any ref in a throwaway worktree.

Breaking changes:
- Output is a `.tar.bz2` tarball + `.yml` manifest, not loose binaries
- Default output dir is `./dist/`, not `./docker-output/`
- Curated binary set: `agave-validator`, `agave-watchtower`, `agave-ledger-tool`, `solana-genesis`, `solana-gossip`, `solana-faucet`, `solana`, `solana-keygen`
- Requires `docker buildx` (Docker Engine >=23 or the buildx plugin)
- CLI is all flags: `./f true` → `./f --profile debug`


0xEdgar (dismissed):

f
```diff
@@ -1,30 +1,282 @@
...
+
+if [[ -n "$platform" ]]; then
+  buildx_args+=(--platform "$platform")
+fi
+
+if [[ "$no_cache" -eq 1 ]]; then
+  buildx_args+=(--no-cache)
+fi
+
+for ref in "${pull_cache_refs[@]}"; do
```
0xEdgar, line 305: fable caught this: 

Empty-array expansion under set -u is an "unbound variable" error on bash < 4.4 — including macOS stock /bin/bash 3.2, which #!/usr/bin/env bash resolves to unless brew bash is installed. Since pull_cache_refs/push_cache_refs are empty by default, plain ./f dies here before building anything (reproduced on 3.2.57). The old ./f worked on the same machines.

Fix is the standard guard, here and on the push_cache_refs loop below:
```rust
for ref in ${pull_cache_refs[@]+"${pull_cache_refs[@]}"}; do
```
(Alternative: require bash ≥ 4.4 with an explicit version check at the top — but the guard keeps Mac users working.)

0xEdgar (approved):

f
jito-spaceman: Implemented bash version checks in line with our other bash scripts across repos. Minimum req of bash 4, meaning out-of-the-gate 3.2 on MacOS is no longer valid.

PR #1454, opened 2026-05-26
[v4.0] Increase BAM shred retransmit lookahead

esemeniuc:
## Summary
- Increase BAM shred retransmit leader lookahead from 2 slots to 4


haoranjito:
is this a duplicate for #1455?

haoranjito (approved):

PR #1455, opened 2026-05-26
Increase BAM shred retransmit lookahead

esemeniuc:
## Summary
- Increase BAM shred retransmit leader lookahead from 2 slots to 4


haoranjito (approved):
lgtm

PR #1465, opened 2026-06-05
[v4.1] Fix local cluster startup

esemeniuc:
## Summary
- Create ledgers for non-bootstrap BAM local cluster validators before startup.
- Set a short snapshot interval for local cluster validator processes.
- Restore hidden `--gossip-host` / `--advertised-ip` argument support for validator startup.

## Why
Local cluster startup fail when non-bootstrap validators launch without initialized ledgers, and validator startup rejects the gossip host flag passed by the cluster manager.


haoranjito (approved):
lgtm

PR #1467, opened 2026-06-05
Enable TLS Connections to BAM Node

badgerrupt0:
This change enables validators to connect to the a BAM node over HTTPS, and changes the default scheme from HTTP -> HTTPS. This is done by borrowing the same technique used to connect to BE/relayer proxies based off of whether the URL contains `https` or `http`. HTTP is still supported but if left unspecified, HTTPS is used. To switch a validator over from HTTP, one can either change their BAM url in their config and restart, or use the `set-bam-config --bam-url` command to change it.

### DIFF
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -186,7 +186,7 @@ thiserror = { workspace = true }
 tokio = { workspace = true, features = ["full"] }
 tokio-stream = { workspace = true }
 tokio-util = { workspace = true }
-tonic = { workspace = true }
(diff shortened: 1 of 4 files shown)


ebin-mathews:
@badgerrupt0 lgtm. Lets sync on the release

0xEdgar (commented):
pref for keeping default to http and flipping default to https in a future PR, otherwise lgtm.

0xEdgar (approved):

PR #1471, opened 2026-06-13
Fix simulateBundle failure details

### LINKED ISSUE #678: simulate_bundle rpc method does not return full result on transaction error
#### Problem
When calling the simulate_bundle RPC method and one of transactions fails due to e.g. custom thrown error from an instruction, the result does not contain neither logs nor consumed compute units, and the error itself is returned only as string which makes it cumbersome to parse.

example result:
```
Response { context: RpcResponseContext { slot: 290184069, api_version: Some(RpcApiVersion(Version { major: 2, minor: 0, patch: 5 })) }, value: RpcSimulateBundleResult { summary: Failed { error: TransactionFailure(4SHwZHvvMWgL2RarUBrsQpWAkVPRXR1EezULc56Ej7ffjmMrwAEG2Zd1cuQvCM45t4eVv1xT92JKnP3srjarXrbQ, "Error processing Instruction 3: custom program error: 0x1772"), tx_signature: Some("<REDACTED>") }, transaction_results: [] } }
```

As you can see, only `summary` field is returned; the `transaction_results` is empty.

#### Proposed Solution
`transaction_results` should be returned; It has all the fields that standard solana `simulate_transaction` rpc method returns: logs, consumed compute units etc. This helps to debug failures.

beriktassuly:
## Summary
- Preserve execution details for failed transactions during bundle simulation.
- Keep bundle simulation all-or-nothing behavior while returning logs and consumed units for the failed transaction.
- Add RPC regression assertions for failed transaction logs and consumed units.

## Tests
- Not run locally.

Fixes #678


Berektassuly:
Verified locally with the Windows unpack fix from #1473 applied:

- `cargo test -p solana-rpc-test test_simulate_bundle --test rpc -- --nocapture`
- `cargo test -p solana-runtime test_simulate_transactions_unchecked_with_pre_accounts -- --nocapture`

Both passed.

PR #1473, opened 2026-06-14
Open unpack directories on Windows

beriktassuly:
#### Problem

On native Windows, genesis archive verification can fail while unpacking with:

```text
Error checking to unpack genesis archive: IO error: Access is denied. (os error 5)
```

The hardened unpacker opens the destination parent directory with `File::open(parent)`. Windows requires directory handles to be opened with `FILE_FLAG_BACKUP_SEMANTICS`, so the normal file-open path can fail before the archive contents are verified.

#### Summary of Changes

- Add a small `open_dir()` helper that preserves the current non-Windows behavior.
- Open unpack destination directories with `FILE_FLAG_BACKUP_SEMANTICS` on Windows.
- Add a focused test that verifies the helper can open a directory handle.

#### Tests

- `cargo +nightly fmt --check -- snapshots/src/hardened_unpack.rs`
- `cargo test -p agave-snapshots --features agave-unstable-api test_open_dir_opens_directory -- --nocapture`
- `cargo test -p agave-snapshots --features agave-unstable-api unpack_genesis -- --nocapture`
- `cargo test -p solana-rpc-test test_simulate_bundle --test rpc -- --nocapture`
- `cargo test -p solana-runtime test_simulate_transactions_unchecked_with_pre_accounts -- --nocapture`

### DIFF
--- a/snapshots/src/hardened_unpack.rs
+++ b/snapshots/src/hardened_unpack.rs
@@ -230,6 +230,27 @@ fn should_fallback_to_tar_unpack<R: io::Read>(entry: &tar::Entry<'_, R>) -> bool
     ) || entry.header().as_ustar().is_none() && entry.path_bytes().ends_with(b"/")
 }
 
+fn open_dir(path: &Path) -> io::Result<File> {
+    open_dir_impl(path)
+}
+
+#[cfg(not(windows))]
+fn open_dir_impl(path: &Path) -> io::Result<File> {
+    File::open(path)
+}
+
+#[cfg(windows)]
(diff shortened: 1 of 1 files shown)


esemeniuc:
Sorry we don't test or develop on windows

esemeniuc:
Snapshot handling is upstream issue, please propose your change in agave repo. we will pull in changes from there https://github.com/anza-xyz/agave/

Berektassuly:
Thanks for the clarification! I've already opened the upstream PR in Agave: anza-xyz/agave#13190

PR #1478, opened 2026-06-26
feat: support txv1 transactions in bundle and BAM ingest

### LINKED ISSUE #1452: Add support for txv1 for bundles and BAM packets/bundles
#### Problem
v4.2 introduces a larger transaction format, txv1. We do not support these in BAM and bundles because we truncate packets to the previous 1232 byte max payload.

#### Proposed Solution
Add support for txv1

bit2swaz:
Fixes #1452

## What's broken

Protocol v4.2 added txv1 transactions which can be up to 4096 bytes. The old packet size limit was 1232 bytes (`PACKET_DATA_SIZE`). Two converter functions, i.e. one on the relayer/block-engine path (`core/src/lib.rs`) and one on the BAM path (`bam_receive_and_buffer.rs`) both capped their byte copies at that old limit. Any txv1 transaction over 1232 bytes was silently truncated before it
reached the sanitizer, while on the BAM path it was outright discarded.

The sanitizer already handles txv1 correctly. The converters just had a stale ceiling

## What changed

Raised the copy ceiling in both converters from `PACKET_DATA_SIZE` (1232) to `solana_message::v1::MAX_TRANSACTION_SIZE` (4096). Legacy and v0 behavior is unchanged since the sanitizer still enforces the old limit for those formats.

The BAM converter was also hoisted from a nested function inside `batch_verify` to `impl` scope so it can be tested directly. The oversized-packet discard path was cleaned up to skip the heap allocation entirely.

**core/src/lib.rs** (relayer/block-engine path, called from `relayer_stage.rs` and `block_engine_stage.rs`):
- one-line ceiling change
- 3 unit tests covering txv1-sized, over-max, and legacy packet sizes

**core/src/banking_stage/transaction_scheduler/bam_receive_and_buffer.rs** (BAM path):
- ceiling raised, converter moved to impl scope
- 2 direct converter tests
- 1 end-to-end test using a real signed txv1 transaction that exceeds 1232 bytes

## Test results

```
cargo test -p solana-core -- proto_packet bam_receive_and_buffer proxy bundle
test result: ok. 56 passed; 0 failed
```

### DIFF
--- a/core/src/banking_stage/transaction_scheduler/bam_receive_and_buffer.rs
+++ b/core/src/banking_stage/transaction_scheduler/bam_receive_and_buffer.rs
@@ -41,7 +41,7 @@ use {
     solana_accounts_db::account_locks::validate_account_locks,
     solana_clock::{MAX_PROCESSING_AGE, Slot},
     solana_measure::{measure::Measure, measure_us},
-    solana_packet::{Meta, PACKET_DATA_SIZE, PacketFlags},
+    solana_packet::{Meta, PacketFlags},
     solana_perf::{
         packet::{BytesPacket, BytesPacketBatch},
         sigverify::ed25519_verify,
@@ -649,6 +649,29 @@ impl BamReceiveAndBuffer {
         stats
     }
 
+    fn proto_packet_to_packet(from_packet: Packet) -> BytesPacket {
+        let Packet { data, meta } = from_packet;
+        let data_len = data.len();
+        if data_len > solana_message::v1::MAX_TRANSACTION_SIZE {
+            let mut p = BytesPacket::new(Bytes::new(), Meta::default());
+            p.meta_mut().set_discard(true);
+            return p;
+        }
+
+        let mut to_packet = BytesPacket::new(data.into(), Meta::default());
+        to_packet.meta_mut().size = data_len;
+
+        if let Some(meta) = meta
+            && meta.flags.is_some_and(|flags| flags.simple_vote_tx)
+        {
+            to_packet
+                .meta_mut()
+                .flags
+                .insert(PacketFlags::SIMPLE_VOTE_TX);
+        }
+        to_packet
+    }
+
     fn batch_verify(
         sigverify_thread_pool: &rayon::ThreadPool,
         atomic_txn_batches: &mut [MultipleAtomicTxnBatch],
@@ -672,28 +695,7 @@ impl BamReceiveAndBuffer {
                 let solana_packet_batch: Vec<_> = atomic_txn_batch
                     .packets
                     .drain(..)
-                    .map(|from_packet| {
-                        let Packet { data, meta } = from_packet;
-                        let data_len = data.len();
-                        if data_len > PACKET_DATA_SIZE {
-                            let mut to_packet = BytesPacket::new(Bytes::new(), Meta::default());
-                            to_packet.meta_mut().set_discard(true);
-                            return to_packet;
-                        }
-
-                        let mut to_packet = BytesPacket::new(data, Meta::default());
-                        to_packet.meta_mut().size = data_len;
-
-                        if let Some(meta) = meta
-                            && meta.flags.is_some_and(|flags| flags.simple_vote_tx)
-                        {
-                            to_packet
-                                .meta_mut()
-                                .flags
-                                .insert(PacketFlags::SIMPLE_VOTE_TX);
-                        }
-                        to_packet
-                    })
+                    .map(Self::proto_packet_to_packet)
                     .collect();
                 packet_count += solana_packet_batch.len();
                 packet_batches.push(solana_perf::packet::PacketBatch::Bytes(
@@ -1315,7 +1317,7 @@ mod tests {
         let batch = AtomicTxnBatch {
             seq_id: 1,
             packets: vec![Packet {
-                data: vec![0; PACKET_DATA_SIZE + 1].into(),
+                data: vec![0; solana_message::v1::MAX_TRANSACTION_SIZE + 1].into(),
                 meta: None,
             }],
             max_schedule_slot: Slot::MAX,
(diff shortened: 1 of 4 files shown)


bit2swaz:
just three things i wanted to confirm before this got merged. none of them block the change, i just wanted to make sure i havent missed anything

**1. feature gating:**
should the wider acceptance window (1232 to 4096 bytes) be gated behind the txv1 feature flag, or is it fine to accept larger packets unconditionally even before txv1 is active on a cluster? the constant resolves today, but i wasnt sure whether acceptance policy should track the runtime feature gate

**2. meta.size trust on the relayer path:**
the `lib.rs` converter copies `meta.size` directly from the proto payload. with packets now up to 4096 bytes, is there any concern about a malicious relayer over reporting that field? my take is that sigverify and the sanitizer already close the door here but i wanted a second pair of eyes on this

**3. Sigverify batch sizing on the BAM path:**
does `batch_verify` or `bundle_sigverify_stage` have any hidden per packet size assumption for SIMD batching? i couldnt find one while tracing the code but it seemed worth confirming since the max packet size is now ~3.3x larger

bit2swaz:
@buffalu could you review this whenever youve got the time?

bit2swaz:
hey @buffalu, this has been open for a month, would appreciate a review whenever you get a chance. 

also noting that upstream's recent "Jito Patch" (fd31d7b) reverted `proto_packet_to_packet` back to `PACKET_DATA_SIZE` which is exactly the truncation bug this PR fixes. might be worth keeping that in mind during review :)

core/src/lib.rs
```diff
@@ -87,7 +87,7 @@ const UNKNOWN_IP: IpAddr = IpAddr::V4(Ipv4Addr::new(0, 0, 0, 0));
 // NOTE: last profiled at around 180ns
 pub fn proto_packet_to_packet(p: jito_protos::proto::packet::Packet) -> BytesPacket {
     let mut data = p.data;
-    data.truncate(PACKET_DATA_SIZE);
+    data.truncate(solana_message::v1::MAX_TRANSACTION_SIZE);
```
0xEdgar: Could we avoid truncating here? Should just reject above 4096

core/src/banking_stage/transaction_scheduler/bam_receive_and_buffer.rs
0xEdgar: we'll need to thread through enable_tx_v1 as an argument here

core/src/banking_stage/transaction_scheduler/bam_receive_and_buffer.rs
```diff
@@ -1622,4 +1624,83 @@ mod tests {
...
+                    .map(|_| {
+                        system_instruction::transfer(&payer_pubkey, &Pubkey::new_unique(), 1)
+                    })
+                    .collect();
+                let message =
+                    v1::Message::try_compile(&payer_pubkey, &ixs, recent_blockhash).unwrap();
+                let tx =
+                    VersionedTransaction::try_new(VersionedMessage::V1(message), &[&mint_keypair])
+                        .unwrap();
+                bincode::serialize(&tx).unwrap()
```
0xEdgar: txv1 uses wincode instead of bincode

0xEdgar (commented):
Hey! thanks for putting this together, and apologies it took us this long to give you a  review. We were waiting for txv1 wire format to stabilize before working on this. 

The overall direction is solid, left some inline comments in the review that need addressing. 

Also my LLM found an additional sigverify path outside the current diff: BundleSigverifyStage also passes enable_tx_v1 = false, so V1 bundles would still be rejected after activation. It should receive the working-bank state as an arg. 

We’re hoping to get this landed this week. Since maintainer edits are enabled, we’re happy to start adding the follow-up commits to this branch while preserving your original commit and authorship. Let us know your expected timeline so we can avoid duplicating work. Otherwise, we’ll plan to build on what you have here and carry it through.

bit2swaz:
got it on all four. heres what im thinking:

> Could we avoid truncating here? Should just reject above 4096

yep. the BAM path already does this with `set_discard(true)` + empty buffer. ill mirror that in [lib.rs#L88-L90](https://github.com/jito-foundation/jito-solana/blob/ed6162fbe5484914f9e1b7ff813b8bfc42472d94/core/src/lib.rs#L88-L90) and update the test to assert `meta().discard()` instead of checking data length

> we'll need to thread through enable_tx_v1 as an argument here

the hardcoded `false` at [bam_receive_and_buffer.rs#L707-L713](https://github.com/jito-foundation/jito-solana/blob/ed6162fbe5484914f9e1b7ff813b8bfc42472d94/core/src/banking_stage/transaction_scheduler/bam_receive_and_buffer.rs#L707-L713) drops all v1 packets regardless of cluster state. `run_parsing` already has `bank_forks` in scope and reads the working bank nearby, so ill pull `enable_tx_v1` from the same read and thread it into `batch_verify` as a new bool param

> txv1 uses wincode instead of bincode

[line 1673](https://github.com/jito-foundation/jito-solana/blob/ed6162fbe5484914f9e1b7ff813b8bfc42472d94/core/src/banking_stage/transaction_scheduler/bam_receive_and_buffer.rs#L1673) uses `bincode::serialize` but v1 puts the message body first with no signature length prefix which bincode doesnt produce. switching to `wincode::serialize` (already a `core` dep) matches what `sigverify_stage.rs` does for its txv1 test

re: `BundleSigverifyStage`, same fix there too. ill add `bank_forks` to `new()` and read the flag per iteration, then update the three unit tests that call `new()` directly

planning to have everything up before friday. can you hold off on pushing to the branch until then?

haoranjito:
@bit2swaz  This branch has been cherrypicked into https://github.com/jito-foundation/jito-solana/pull/1552

Thanks for the original tx-v1 bam and bundle sigverify work!

PR #1480, opened 2026-06-28
Polling cleanup

esemeniuc:
#### Problem

BAM outbound handling used a dedicated polling task over a clonable receiver, waking frequently even when there was no outbound work and complicating reconnect ownership.

#### Summary of Changes

- Move BAM outbound handling to a single owned Tokio `mpsc` receiver inside the connection task.
- Preserve the outbound receiver across reconnects with explicit connection shutdown.
- Keep batching for queued transaction batch results while handling leader state sends inline.
- Remove unused outbound message variants and ping/pong-only metrics.
- Respond to BAM pings directly from the inbound scheduler stream.

#### Performance

- Eliminates the 1ms outbound polling task and its periodic wakeups.
- Uses async receiver wakeups instead of repeatedly draining an idle channel.
- Reduces task/channel overhead on the BAM connection path.

### DIFF
--- a/core/tests/bam_connection.rs
+++ b/core/tests/bam_connection.rs
@@ -9,8 +9,8 @@ use {
             scheduler_response_v0::Resp,
         },
         bam_types::{
-            AtomicTxnBatch, BamConfig, BlockEngineBuilderConfig, BuilderHeartBeat,
-            MultipleAtomicTxnBatch, Packet, Socket,
+            AtomicTxnBatch, AtomicTxnBatchResult, BamConfig, BlockEngineBuilderConfig,
+            BuilderHeartBeat, MultipleAtomicTxnBatch, Packet, Socket,
         },
     },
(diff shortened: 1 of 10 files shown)


core/tests/bam_connection.rs
```diff
@@ -194,22 +194,39 @@ impl BamNodeApi for MockBamNode {
...
-
-                let batch = batch_to_send.lock().unwrap().take();
-                if let Some(batch) = batch {
-                    let resp = v0_response(Resp::MultipleAtomicTxnBatch(MultipleAtomicTxnBatch {
-                        batches: vec![batch],
-                    }));
-                    if tx.send(Ok(resp)).await.is_err() {
-                        break;
+            loop {
+                tokio::select! {
```
haoranjito, line 202: 👍

haoranjito (approved):

PR #1483, opened 2026-06-29
[codex] Run cargo audit at the end of CI

esemeniuc:
## Summary
- Remove `cargo audit` from the early `ci/test-checks.sh` check path.
- Add a dedicated generated Buildkite `cargo audit` step after the final test stage for public, private, PR, and merge queue pipelines.
- Move the secondary pipeline `cargo audit` job after publish jobs.

## Why
RustSec checks should still fail CI, but only after the rest of the Buildkite work has had a chance to run.

### DIFF
--- a/ci/buildkite-secondary.yml
+++ b/ci/buildkite-secondary.yml
@@ -2,13 +2,6 @@
 # Build steps that run after the primary pipeline on pushes and tags.
 # Pull requests to not run these steps.
 steps:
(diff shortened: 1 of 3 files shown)


0xEdgar (changes requested):
think we should still have this fail; we can reorder the run so that this surfaces after the more important checks run first

0xEdgar (approved):

PR #1487, opened 2026-06-30
BAM URL switch cleanups

esemeniuc:
#### Problem

BAM URL changes were only logged when an active connection noticed a URL mismatch. If the validator was already disconnected or retrying, a later URL change could be picked up silently.

During intentional BAM disconnects, `refresh_config` could also be waiting on its interval or config RPC. Shutdown only waited 100ms before aborting it, causing this noisy warning:

`BAM task refresh_config did not exit within 100ms; aborting`

#### Summary of Changes

- Track the last observed BAM URL in `BamManager` and log every configured URL change.
- Cancel the BAM config refresh task with a one-shot shutdown signal before waiting on task shutdown.
- Remove the redundant `set_bam_url` debug log.
- Add coverage for shutdown while config refresh is in flight

### DIFF
--- a/core/src/bam_connection.rs
+++ b/core/src/bam_connection.rs
@@ -25,7 +25,7 @@ use {
     },
     thiserror::Error,
     tokio::{
-        sync::mpsc,
+        sync::{mpsc, oneshot},
         time::{interval, timeout},
     },
     tokio_stream::wrappers::ReceiverStream,
@@ -156,6 +156,7 @@ impl BamConnection {
         }
 
         let mut post_auth_client = Some(validator_client);
+        let mut refresh_config_exit_sender = None;
         let mut refresh_config_task = None;
 
         while !connection_exit.load(Relaxed) {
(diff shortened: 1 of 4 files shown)


haoranjito (dismissed):
:shipit:

haoranjito:
may need to run cargo fmt


+ cargo +nightly-2026-02-28 fmt --all -- --check
--
  | Diff in /solana/core/tests/bam_connection.rs:509:
  | async fn test_shutdown_cancels_inflight_config_refresh() {
  | let send_heartbeats = Arc::new(AtomicBool::new(true));
  | let server = start_mock_server(send_heartbeats, Duration::from_secs(1), false).await;
  | -        server.config_response_delay_ms.store(5_000, Ordering::Relaxed);
  | +        server
  | +            .config_response_delay_ms
  | +            .store(5_000, Ordering::Relaxed);

haoranjito (approved):

PR #1499, opened 2026-07-03
Centralize Tonic Endpoint Creation

badgerrupt0:
#### Problem

A previous issue patched by #1493 was created when a tonic upgrade changed the default behavior when it came to loading in trust roots. The callsite for the BE proxy connection was not properly updated with this change, so trust roots were never loaded in and created issues connecting directly to the block engine. This also revealed a gap in testing for the endpoint creation.

#### Summary of Changes

This change centralizes the creation of tonic endpoints for both BAM and BE so that any future changes to tonic will apply to both, and adds  relevant additional test coverage.


0xEdgar (commented):
can you add a handshake test through the production builder(s), using the generated CA via SSL_CERT_FILE? Want to ensure test suite proves can catch regressions like last week. 

```rust
let endpoint = endpoint_from_url_with_tls_config(
    &url,
    ClientTlsConfig::new().ca_certificate(Certificate::from_pem(cert_pem)),
)
```

core/src/tonic_endpoint.rs
```diff
@@ -155,4 +162,31 @@ mod tests {
             "TLS connection unexpectedly succeeded without configured roots"
         );
     }
+
+    #[tokio::test]
+    async fn endpoint_handshakes_with_testnet_block_engine() {
+        const TESTNET_BLOCK_ENGINE_URL: &str = "https://testnet.block-engine.jito.wtf";
+        const CONNECT_TIMEOUT: Duration = Duration::from_secs(10);
+
+        let endpoint = endpoint_from_url(TESTNET_BLOCK_ENGINE_URL).unwrap();
```
0xEdgar: doesn't BE use `endpoint_from_url_with_error` ?

badgerrupt0: The endpoint construction logic is the same, just with supplied error callbacks, but better to keep the exact same I agree- updated now!

0xEdgar (approved):

PR #1504, opened 2026-07-09
Cleanup BAM Connection

esemeniuc:
#### Problem

BAM connection and scheduler paths did extra per-batch channel/metric work, and BAM manager reapplied builder config state even when the config had not changed.

#### Summary of Changes

- Send `MultipleAtomicTxnBatch` messages to scheduler instead of AtomicTxnBatch, reduces lock acquisition cost for channel
- Track builder config versions so BAM manager only reapplies TPU, shred socket, and block builder state when a new config arrives. Currently happens ~200x per second
- Keep BAM connection metrics as task-local counters where cross-task atomics are not needed.

### DIFF
--- a/bam-banking-bench/src/mock_bam_server.rs
+++ b/bam-banking-bench/src/mock_bam_server.rs
@@ -1,7 +1,9 @@
 use {
     bytes::{BufMut, BytesMut},
     crossbeam_channel::Sender,
-    jito_protos::proto::bam_types::{AtomicTxnBatch, AtomicTxnBatchResult, Packet},
+    jito_protos::proto::bam_types::{
+        AtomicTxnBatch, AtomicTxnBatchResult, MultipleAtomicTxnBatch, Packet,
+    },
     solana_compute_budget_interface::ComputeBudgetInstruction,
     solana_core::bam_dependencies::BamOutboundMessage,
     solana_hash::Hash,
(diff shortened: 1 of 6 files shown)


ebin-mathews:
```
Send inbound MultipleAtomicTxnBatch messages to scheduler as grouped payloads instead of one channel send per atomic batch (avoid atomic updates)
```

Might make latencies worse

esemeniuc:
> ```
> Send inbound MultipleAtomicTxnBatch messages to scheduler as grouped payloads instead of one channel send per atomic batch (avoid atomic updates)
> ```
> 
> Might make latencies worse

confused how would it make it worse? rn we ungroup to send in channel then to reaggregate?

ebin-mathews:
> > ```
> > Send inbound MultipleAtomicTxnBatch messages to scheduler as grouped payloads instead of one channel send per atomic batch (avoid atomic updates)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Might make latencies worse
> 
> confused how would it make it worse? rn we ungroup to send in channel then to reaggregate?

Maybe the PR description should say so then?

Something like this?
```
one channel send per atomic batch instead of ungroup, send in channel and re-aggregate.
```

esemeniuc:
> > > ```
> > > Send inbound MultipleAtomicTxnBatch messages to scheduler as grouped payloads instead of one channel send per atomic batch (avoid atomic updates)
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Might make latencies worse
> > 
> > 
> > confused how would it make it worse? rn we ungroup to send in channel then to reaggregate?
> 
> Maybe the PR description should say so then?
> 
> Something like this?
> 
> ```
> one channel send per atomic batch instead of ungroup, send in channel and re-aggregate.
> ```

changed verbiage

haoranjito (approved):
lgtm.

PR #1519, opened 2026-07-16
Report missing nonce authority as instruction error in BAM

esemeniuc:
## Summary

Report a missing durable-nonce authority signature as an instruction error in the BAM completion path.

| Before | After |
| --- | --- |
| A durable-nonce transaction with a valid nonce account and matching nonce, but no authority signature, was reported as `BLOCKHASH_NOT_FOUND`. | The same failure is reported as `INSTRUCTION_ERROR` with `MissingRequiredSignature`. |
| Genuine missing, stale, mismatched, or non-advanceable nonce failures were reported as `BLOCKHASH_NOT_FOUND`. | Those failures remain `BLOCKHASH_NOT_FOUND`. |

## Why

`BLOCKHASH_NOT_FOUND` suggests that the nonce is missing or invalid. In this case the nonce is valid—the transaction is missing authorization. Reporting `MissingRequiredSignature` identifies the actual problem and tells callers what they need to fix.

The reclassification is deliberately narrow and only happens when the transaction and bank state confirm this exact case.

### DIFF
--- a/core/src/banking_stage/consume_worker.rs
+++ b/core/src/banking_stage/consume_worker.rs
@@ -8,7 +8,10 @@ use {
             TransactionResult,
         },
     },
-    crate::banking_stage::consumer::{ExecutionFlags, RetryableIndex, TipProcessingDependencies},
+    crate::banking_stage::{
+        consumer::{ExecutionFlags, RetryableIndex, TipProcessingDependencies},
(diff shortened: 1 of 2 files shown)


0xEdgar (approved):

haoranjito (approved):
lgtm

esemeniuc:
Closing, taking alternative approach

PR #1533, opened 2026-08-05
ci: bind workflow inputs via env before shell

SashaMIT:
## Summary
Bind workflow `inputs.*` / event context into step `env:` before shell use in:
- `backport.yml` (`pull_number` / PR metadata)
- `bump-version.yml` (`level`)

Keeps `${{ }}` expansions out of `run:` scripts.

## Test plan
- [ ] Confirm workflow YAML still validates
- [ ] Optional: dry-run backport / bump-version dispatch paths

Made with [Cursor](https://cursor.com)


esemeniuc:
what is this for?

SashaMIT:
Good question — it's CI hardening in the class of GitHub's script-injection guidance (https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#understanding-the-risk-of-script-injections): values like workflow inputs should be bound through `env:` rather than expanded inline in `run:` blocks, so a maliciously crafted value can't break out into shell.

Fair note on these two specific inputs (from the same discussion on anza-xyz/agave#14336): `bump-version.level` is `type: choice` (hardcoded list), so there's no live injection today — this change is uniformity / future-proofing so the dangerous pattern isn't there if the input type ever loosens. The `backport.yml` side binds PR metadata (author/title-derived values), which is attacker-influenced text and is the part that carries real risk.

Happy to split so the backport half lands on its own merits, or narrow further if you'd prefer only attacker-controlled fields bound.

PR #1536, opened 2026-08-12
Fix bundle entry byte accounting

### LINKED ISSUE #1421: Merge the entry_bytes changes into master for both normal consumer and bundle consumer

esemeniuc:
## Summary

- reserve serialized entry bytes for every transaction recorded by `BundleConsumer`
- include per-transaction entry overhead because `record_bundle()` creates one entry per transaction
- reject over-budget bundles as `PohRecorderError::MaxHeightReached` before conversion, recording, or commit
- preserve immediate retry behavior and QoS cost release
- add five-transaction one-byte-short and exact-boundary regression coverage

## Root cause

The ordinary transaction consumer reserves from the bank's shared entry-byte budget before recording, but the bundle path called `record_bundle()` without making the equivalent reservation. Bundle entries could therefore exceed the slot's serialized entry-byte limit.

## Impact

Bundle and ordinary transaction consumers now coordinate through the same atomic per-slot budget. An over-budget bundle is rejected before reaching PoH or mutating bank state, while an exact-boundary bundle remains recordable.

Closes #1421.

### DIFF
--- a/core/src/bundle_stage/bundle_consumer.rs
+++ b/core/src/bundle_stage/bundle_consumer.rs
@@ -2,8 +2,8 @@ use {
     crate::banking_stage::{
         committer::{CommitTransactionDetails, Committer},
         consumer::{
-            ExecuteAndCommitTransactionsOutput, ExecutionFlags, LeaderProcessedTransactionCounts,
-            ProcessTransactionBatchOutput, RetryableIndex,
+            ENTRY_OVERHEAD_BYTES, ExecuteAndCommitTransactionsOutput, ExecutionFlags,
+            LeaderProcessedTransactionCounts, ProcessTransactionBatchOutput, RetryableIndex,
         },
         leader_slot_timing_metrics::LeaderExecuteAndCommitTimings,
         qos_service::QosService,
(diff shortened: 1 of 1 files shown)


core/src/bundle_stage/bundle_consumer.rs
```diff
@@ -388,46 +392,59 @@ impl BundleConsumer {
...
+                        .sanitized_transactions()
+                        .iter()
+                        .map(|tx| tx.to_versioned_transaction())
+                        .collect_vec()
+                );
+
+                // BundleStage executes multiple transactions which may contain overlapping
+                // accounts. This is needed until relax_intrabatch_account_locks is enabled.
+                let (summary, record_us) = measure_us!(
+                    self.transaction_recorder
```
buffalu, line 425: for future self, we can remove this separate bundle recording and do in a giant batch now that simd 83 is merged

buffalu (approved):

PR #1537, opened 2026-08-12
Fix BAM Block Engine cutover

esemeniuc:
## Summary

- coordinate the transition from Block Engine bundle processing to BAM through the existing shared state
- clear buffered Block Engine bundles before BAM configuration becomes active
- acknowledge the drain before publishing BAM as connected
- use acquire reads for every production consumer of the published BAM state
- prevent bundles received while BAM is active from executing after a later disconnect

## Root cause

BAM could publish the connected state while BundleStage still retained bundles accepted through the Block Engine path. Those bundles could execute after the BAM path became active.

## Impact

The handoff is now ordered as Connecting -> Draining Block Engine -> Block Engine Drained -> Connected. The release/acquire chain also publishes BAM configuration consistently to state-dependent consumers.

### DIFF
--- a/core/src/bam_manager.rs
+++ b/core/src/bam_manager.rs
@@ -3,7 +3,7 @@
 /// - Sends leader state to BAM
 /// - Updates TPU config
 /// - Updates block builder fee info
-/// - Sets `bam_enabled` flag that is used everywhere
+/// - Coordinates the switch between Block Engine bundle processing and BAM
 use std::{
     net::{Ipv4Addr, SocketAddr, SocketAddrV4},
     str::FromStr,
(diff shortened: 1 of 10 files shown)


core/src/bam_manager.rs
haoranjito, line 264: P1 core/src/bam_manager.rs:261: after entering DrainingBlockEngine, the manager still sends LeaderState before bam_enabled is Connected. The BAM receive path drops batches unless the state is exactly Connected (core/src/banking_stage/transaction_scheduler/                    
    bam_receive_and_buffer.rs:743 and :764), so BAM can respond to these early leader-state messages with batches that are immediately drained/dropped during the handoff window. Gate leader-state sends on the post-transition Connected state, or delay them until after config is   
    applied and bam_enabled is stored as Connected.

esemeniuc: updated

haoranjito (approved):

PR #1539, opened 2026-08-12
Simplify bundle recording after SIMD-0083

esemeniuc:
Follow-up to [comment](https://github.com/jito-foundation/jito-solana/pull/1536#discussion_r3767186584) in  #1536.

## Summary

- Record each bundle as a single transaction entry.
- Convert bundle transactions once and calculate exact single-entry bytes before taking the freeze lock.
- Remove the Jito-only batched recorder APIs and redundant SIMD-0083 locking wrapper.
- Remove stale bundle execution plumbing, dead calculations, and obsolete tests.

## Motivation

Bundle recording retained a pre-SIMD-0083 workaround that emitted one entry per transaction so transactions with overlapping accounts could coexist in a bundle. SIMD-0083 now permits intrabatch account conflicts and preserves transaction order, so the separate entries and special locking path are no longer necessary.

## Impact

- Reduces bundle record-channel insertions and PoH mixins from as many as five to one.
- Saves 48 bytes of entry overhead for every bundle transaction after the first.
- Avoids converting successful bundle transactions twice and shortens the freeze-lock critical section.
- Removes 215 net lines from the Jito patch.
- Preserves bundle order, all-or-nothing admission, retry behavior, and transaction indexes.

### DIFF
--- a/core/src/banking_stage/consume_worker.rs
+++ b/core/src/banking_stage/consume_worker.rs
@@ -3248,7 +3248,6 @@ mod tests {
     }
 
     fn setup_test_frame(
-        _relax_intrabatch_account_locks: bool,
         default_rent: bool,
     ) -> (
         TestFrame,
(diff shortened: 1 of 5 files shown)


haoranjito (approved):
lgtm.

PR #1541, opened 2026-08-12
Resanitize bundled transactions against execution bank

esemeniuc:
## Summary

Buffered bundle transactions now receive the same execution-Bank validation as other aged transactions before cost reservation, execution, or recording. This prevents invalid vote-only blocks and keeps feature-transition validation consistent across transaction consumers.

- Reuse `Bank::resanitize_transaction_minimally` when BundleStage consumes buffered transactions.
- Reject ordinary bundled transactions when the execution Bank is vote-only.
- Revalidate the feature-gated instruction-account limit after an epoch transition.
- Add regression coverage for both cases.

## Root cause

BundleConsumer maintained a copied subset of the Bank's minimal re-sanitization checks. The copy handled reserved keys and address lookup table invalidation, but did not inherit later vote-only and instruction-account checks added to the shared Bank helper.

A transaction buffered against an earlier Bank could therefore be consumed against a Bank with different validation requirements. In particular, an ordinary transaction could be recorded by a vote-only leader and then rejected during replay with `UserTransactionsInVoteOnlyBank`.

### DIFF
--- a/core/src/bundle_stage/bundle_consumer.rs
+++ b/core/src/bundle_stage/bundle_consumer.rs
@@ -10,6 +10,7 @@ use {
         scheduler_messages::MaxAge,
     },
     itertools::Itertools,
+    smallvec::SmallVec,
     solana_clock::MAX_PROCESSING_AGE,
     solana_measure::measure_us,
     solana_poh::{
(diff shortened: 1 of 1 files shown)


haoranjito (approved):
lgtm.

PR #1542, opened 2026-08-12
Resolve bundle ALTs against the root bank

esemeniuc:
## Issue

BundleStage resolved address lookup tables while buffering against the speculative working bank and cached the resulting runtime transaction. If the working fork changed, the cached addresses could be executed and recorded even though the transaction could not be independently resolved from rooted state.

## Change

Bundle packets now fail buffering when their address lookup tables are unavailable from the rooted bank, preventing a leader from recording a versioned transaction whose resolved addresses cannot be reproduced after a fork change. Sharing the existing packet handler also reduces the maintained Jito-specific validation path.

## Summary

- resolve versioned bundle transactions and reserved account keys against the rooted bank
- retain working-bank vote-only, feature, and cost policy
- reuse the BankingStage packet handler and remove the duplicate BundleStage deserializer
- add regression tests for speculative-only ALT state and working-bank vote-only behavior


0xEdgar (dismissed):

core/src/banking_stage/transaction_scheduler/receive_and_buffer.rs
```diff
@@ -471,7 +472,8 @@ impl TransactionViewReceiveAndBuffer {
 /// ALT deactivation, if any. If no minimum slot, Slot::MAX is returned.
 pub(crate) fn translate_to_runtime_view<D: TransactionData>(
     data: D,
-    bank: &Bank,
+    resolution_bank: &Bank,
```
haoranjito: why do we need to change the signature of this fn? 
seems we can keep the upstream code and avoid this change.

esemeniuc: we need to use the rooted bank for bank operations, and working bank for vote check

esemeniuc: thats the main purpose of this pr. if we didnt change signature, we'd have the old behaviour

haoranjito (approved):

PR #1543, opened 2026-08-12
Avoid blocking the async runtime while waiting for validator startup

questfever:
#### Problem

`wait_for_validator_startup()` performs asynchronous admin RPC calls, but uses `std::thread::sleep()` between connection retries and startup progress checks.

Because the function is driven through `Runtime::block_on()`, these sleeps block the thread polling the future instead of yielding to the Tokio runtime. This can delay other work scheduled on the same runtime while the dashboard waits for the
validator to start.

#### Summary of Changes

Replace the two blocking sleeps inside `wait_for_validator_startup()` with `tokio::time::sleep(...).await`.

The blocking sleep in the synchronous dashboard refresh loop is intentionally left unchanged.

#### Testing

- `git diff --check`
- Attempted `cargo check -p agave-validator --bin agave-validator`

The targeted check progressed through the modified code and dependencies, but could not complete because the local checkout is missing Jito protobuf source files such as `jito-protos/protos/auth.proto`.

Fixes #


questfever:
@ebin-mathews Could you please take a look when you have a chance? Thanks!

PR #1545, opened 2026-08-12
Fail closed atomic transactions on sad handover

esemeniuc:
#### Problem

During an Alpenglow sad leader handover, transactions recorded against an abandoned optimistic parent are re-injected through ordinary ingress. This loses the all-or-nothing semantics of BAM `revert_on_error` batches and BundleStage bundles. If the replacement parent has different state, only a prefix of an atomic batch can be included.

The initial fix prevented protected records from being flattened and re-injected, but exposed a follow-up gap: BAM could already have received a `Committed` result for an atomic batch executed on the optimistic bank even though a later sad handover discarded that bank and neither transaction landed on the selected fork.

#### Summary of Changes

- Add explicit sad-handover rescheduling metadata to each PoH record.
- Mark ordinary transaction records as independently reschedulable.
- Mark BAM `revert_on_error` records and BundleStage records as protected so they are never flattened into ordinary ingress during sad handover.
- Start optimistic leader banks with atomic-batch scheduling disabled.
- Hold BAM atomic batches without executing or responding while the leader bank is provisional.
- Enable atomic batches only after ParentReady resolves the parent:
  - on the happy path, execute them on the confirmed optimistic bank;
  - on sad handover, recreate the leader bank on the ParentReady parent, then execute them atomically on that selected bank.
- Continue rescheduling eligible ordinary transactions across sad handover.
- Extend handover coverage through the BAM scheduler, real consume worker, PoH recorder, and replacement bank.

BAM now reports the atomic result only after the batch has executed against the ParentReady-selected bank. If the replacement fork invalidates any member, the batch returns `NotCommitted` and no valid prefix is committed.

#### Validation

- The sad-handover regression test verifies that an atomic batch receives no result before ParentReady, is retained across a conflicting ParentReady, executes against the replacement bank, returns `NotCommitted` when a fork-only payer is absent, and leaves the otherwise-valid first transaction uncommitted.
- Both live-validator PoCs—the original partial-commit scenario and the follow-up “reported Committed but neither landed” scenario—were run against head `4c72c9e`. In both cases the atomic batch was delivered, but BAM did not report `Committed` during the full pre-ParentReady assertion window.

### DIFF
--- a/core/src/banking_stage/transaction_scheduler/bam_scheduler.rs
+++ b/core/src/banking_stage/transaction_scheduler/bam_scheduler.rs
@@ -63,6 +63,14 @@ fn passthrough_priority(
 
 pub const MAX_PACKETS_PER_BUNDLE: usize = 5; // copied from BundleStorage::MAX_PACKETS_PER_BUNDLE
 
+// Sized from 30 days of mainnet data ending 2026-08-18. Atomic-only ClickHouse counts of distinct
+// `(source, seq_id)` batches per validator-slot had p99=149, p99.9=236, 99.94% <=256, and max=540.
+// The full-slot count conservatively upper-bounds the pre-ParentReady subset stored here.
+// Mainnet-validator Influx `bam_connection-metrics.bundle_received` active 25 ms samples, which
+// also include non-atomic batches, had p99=183, p99.9=272, 99.86% <=256, and max=881. Thus 256
+// keeps the atomic p99.9 inline in 4 KiB while rarer bursts spill safely.
+const DEFERRED_ATOMIC_BATCHES_INLINE_CAPACITY: usize = 256;
+
 pub struct BamScheduler<Tx: TransactionWithMeta> {
     consume_work_sender: Sender<ConsumeWork<Tx>>,
     finished_consume_work_receiver: Receiver<FinishedConsumeWork<Tx>>,
@@ -81,7 +89,8 @@ pub struct BamScheduler<Tx: TransactionWithMeta> {
 
     // Reusable objects to avoid allocations
     reusable_consume_work: Vec<ConsumeWork<Tx>>,
-    deferred_atomic_batches: SmallVec<[TransactionPriorityId; 8]>,
+    deferred_atomic_batches:
+        SmallVec<[TransactionPriorityId; DEFERRED_ATOMIC_BATCHES_INLINE_CAPACITY]>,
 
     extra_checks_enabled: bool,
     bank_forks: Arc<RwLock<BankForks>>,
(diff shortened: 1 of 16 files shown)


core/src/banking_stage/transaction_scheduler/bam_scheduler.rs
```diff
@@ -80,9 +81,11 @@ pub struct BamScheduler<Tx: TransactionWithMeta> {
 
     // Reusable objects to avoid allocations
     reusable_consume_work: Vec<ConsumeWork<Tx>>,
+    deferred_atomic_batches: SmallVec<[TransactionPriorityId; 8]>,
```
haoranjito: why choose 8 here?

haoranjito (commented):
- High: Tip-program upkeep can be lost across sad handover while the slot cache says it ran. BAM non-atomic work can still execute before ParentReady; if it touches tip accounts, maybe_run_tip_programs() records init/crank transactions with revert_on_error = true at core/src/
    banking_stage/consume_worker.rs:247 and marks the slot updated at core/src/banking_stage/consume_worker.rs:296. This PR makes all revert_on_error records non-reschedulable at core/src/banking_stage/consumer.rs:491, and sad handover only re-injects records with
    reschedule_on_sad_handover at core/src/block_creation_loop.rs:1026. Since replay restarts from the UpdateParent boundary, pre-marker upkeep does not carry to the replacement bank. Later BAM work in the same slot skips upkeep because last_tip_updated_slot == bank.slot(). Track
    this cache by (slot, bank_id), reset it on sad handover, or prevent BAM tip upkeep from running until atomic batches are enabled.

  - Medium: make_atomic_consume_or_forward_decision() returns Hold before considering whether there is a working bank at core/src/banking_stage/decision_maker.rs:62. If sad handover clears the optimistic bank and replacement-bank creation fails, clear_bank(true) preserves
    atomic_batches_enabled = false via poh/src/poh_recorder.rs:557, while abort_working_bank() returns early when no bank is attached at core/src/block_creation_loop.rs:862. BundleStage can then hold/buffer bundles outside an active atomic bank until some later reset. Safer
    shape: only gate to Hold when a working bank exists and is provisional, or reset the flag when the provisional bank is abandoned.

core/src/banking_stage/transaction_scheduler/bam_scheduler.rs
esemeniuc: updated

haoranjito:
lgtm. once conflicts are fixed. it should be good to go.

haoranjito (approved):

PR #1546, opened 2026-08-12
Prevent BAM and external scheduler conflicts

esemeniuc:
#### Problem

When an external scheduler takes over BankingStage, it replaces the internal thread set without consuming the BAM receiver. BAM can remain connected while batches queue or drop and unrelated traffic continues through the external scheduler.

#### Summary of Changes

- Reject startup configurations that enable both BAM and external scheduler bindings.
- Reject runtime BAM activation when external scheduler bindings are enabled, while preserving the ability to clear BAM configuration.
- Structurally exclude BAM BankingStage dependencies and the BAM manager whenever external scheduler bindings are enabled.
- Add regression coverage proving that direct runtime mutation of the shared BAM URL cannot activate BAM in external-scheduler mode.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11619,6 +11619,7 @@ dependencies = [
  "solana-transaction",
  "solana-validator-exit",
  "solana-vote-interface",
+ "tempfile",
  "tokio",
 ]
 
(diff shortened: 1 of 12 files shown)


haoranjito (approved):
lgtm.

PR #1551, opened 2026-08-13
Avoid duplicate BAM transaction parsing

esemeniuc:
## Summary
- The PR removes two heap allocations per prevalidated bundle
-- retain the `SanitizedTransactionView<Bytes>` produced during parallel signature verification
- pass the retained view directly into runtime transaction construction
-- remove per-bundle packet adapters, the second `Vec<Bytes>`, byte-buffer clones, and the duplicate parse/sanitize pass
- reuse flat buffers while preserving bundle order and error indices

## Data flow

| Stage | Before | After |
| --- | --- | --- |
| Input | Protobuf `Bytes` | Protobuf `Bytes` |
| Buffering | Allocate `Vec<BytesPacket>` per bundle | Move into one reusable flat `Vec<Bytes>` |
| Verification | In parallel across packets, parse, sanitize, and verify signatures; discard the sanitized view | In parallel across packets, parse, sanitize, and verify signatures; retain the sanitized view |
| Handoff | Allocate `Vec<Bytes>` and clone each `Bytes` handle | Retain `SanitizedTransactionView<Bytes>` |
| Runtime construction | Parse and sanitize again in `parse_batch`, then build `RuntimeTransaction` | Build `RuntimeTransaction` directly from the retained view |

## Impact

This is allocation and architecture cleanup, not a measurable ingest-throughput optimization. It reduces allocation and reference-count churn while preserving transaction-version policy, vote classification, validation ordering, and metrics behavior.

### DIFF
--- a/core/src/banking_stage/transaction_scheduler/bam_receive_and_buffer.rs
+++ b/core/src/banking_stage/transaction_scheduler/bam_receive_and_buffer.rs
@@ -581,8 +581,9 @@ impl BamReceiveAndBuffer {
         atomic_txn_batch_groups: &[MultipleAtomicTxnBatch],
         current_slot: Slot,
         prevalidated: &mut Vec<PrevalidationResult>,
-    ) -> ReceivingStats {
+    ) -> (ReceivingStats, usize) {
         let mut stats = ReceivingStats::default();
+        let mut packet_count = 0;
 
         prevalidated.clear();
         for atomic_txn_batch_group in atomic_txn_batch_groups {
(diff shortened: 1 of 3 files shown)


haoranjito:
---- bundle_stage::tests::test_process_sequential_bundles stdout ----
--
  |  
  | thread 'bundle_stage::tests::test_process_sequential_bundles' (349386) panicked at core/src/bundle_stage.rs:1321:9:
  | assertion `left == right` failed
  | left: 0
  | right: 890880
  |  

<br class="Apple-interchange-newline">

may need to fix the test.

haoranjito (dismissed):
logic looks good to me. please fix the test.

haoranjito (approved):

PR #1552, opened 2026-08-13
feat: support tx v1 in bam receive and buffer

haoranjito:
#### Problem

Credit to @bit2swaz for the core implementation from #1478. I cherry-picked their work here and only adapted it to this branch.

- Add tx-v1 support for bam receive-buffer  
- Add cli to enable tx v1 feature for bam local cluster testing.

#### Testing

Fixes #

### DIFF
--- a/core/src/banking_stage/transaction_scheduler/bam_receive_and_buffer.rs
+++ b/core/src/banking_stage/transaction_scheduler/bam_receive_and_buffer.rs
@@ -1752,9 +1752,7 @@ mod tests {
         let serialized = (1usize..=64)
             .map(|n| {
                 let ixs: Vec<_> = (0..n)
-                    .map(|_| {
-                        system_instruction::transfer(&payer_pubkey, &Pubkey::new_unique(), 1)
-                    })
+                    .map(|_| system_instruction::transfer(&payer_pubkey, &Pubkey::new_unique(), 1))
                     .collect();
                 let message =
                     v1::Message::try_compile(&payer_pubkey, &ixs, recent_blockhash).unwrap();
--- a/core/src/banking_stage/transaction_scheduler/bam_receive_and_buffer.rs
+++ b/core/src/banking_stage/transaction_scheduler/bam_receive_and_buffer.rs
@@ -1205,7 +1205,7 @@ mod tests {
             &thread_pool,
             &mut batches,
             current_slot,
-            false,
+            false, // enable_tx_v1: tests run without the feature active
             metrics,
             &mut prevalidated,
             &mut packet_data,
@@ -1770,7 +1770,7 @@ mod tests {
                 let tx =
                     VersionedTransaction::try_new(VersionedMessage::V1(message), &[&mint_keypair])
                         .unwrap();
-                bincode::serialize(&tx).unwrap()
+                wincode::serialize(&tx).unwrap()
             })
             .find(|bytes| {
                 bytes.len() > solana_packet::PACKET_DATA_SIZE
@@ -1794,9 +1794,9 @@ mod tests {
         let (results, _batch_stats) = run_batch_verify(vec![batch], Slot::MAX, &mut stats);
 
         assert_eq!(results.len(), 1);
-        // batch_verify calls ed25519_verify with enable_tx_v1=false, so a real V1 tx is
-        // rejected at sigverify and surfaces as DeserializationError. That's fine here.
-        // What matters is the packet reaches sigverify at full length, not truncated at copy.
+        // run_batch_verify passes enable_tx_v1=false (feature not active in tests), so a real
+        // V1 tx is rejected at sigverify and surfaces as DeserializationError. That's expected.
+        // What matters is the packet reaches sigverify at full length, not discarded at copy.
         assert!(
             results[0].is_ok() || matches!(&results[0], Err((Reason::DeserializationError(_), _))),
             "txv1 tx must reach sigverify at full length, got unexpected result: {:?}",
--- a/core/src/banking_stage/transaction_scheduler/bam_receive_and_buffer.rs
+++ b/core/src/banking_stage/transaction_scheduler/bam_receive_and_buffer.rs
@@ -1440,7 +1440,7 @@ mod tests {
         let batch = AtomicTxnBatch {
             seq_id: 1,
             packets: vec![Packet {
-                data: vec![0; solana_packet::PACKET_DATA_SIZE + 1].into(),
+                data: vec![0; solana_message::v1::MAX_TRANSACTION_SIZE + 1].into(),
                 meta: None,
             }],
             max_schedule_slot: Slot::MAX,
@@ -1742,5 +1742,65 @@ mod tests {
 
         assert_eq!(results.len(), 1);
         assert!(results[0].is_err());
+    }
+
+    #[test]
+    fn test_txv1_transaction_over_1232_survives_ingest() {
+        use {
+            solana_message::{VersionedMessage, v1},
+            solana_system_interface::instruction as system_instruction,
+        };
+
+        let (bank_forks, mint_keypair) = test_bank_forks();
+        let payer_pubkey = mint_keypair.pubkey();
+        let recent_blockhash = bank_forks.read().unwrap().root_bank().last_blockhash();
+
+        // Pack enough transfer instructions to push the serialized tx over 1232 bytes.
+        // Each extra account key adds ~32 bytes to the message; ~20 instructions clears
+        // PACKET_DATA_SIZE while staying well under MAX_TRANSACTION_SIZE.
+        let serialized = (1usize..=64)
+            .map(|n| {
+                let ixs: Vec<_> = (0..n)
+                    .map(|_| {
+                        system_instruction::transfer(&payer_pubkey, &Pubkey::new_unique(), 1)
+                    })
+                    .collect();
+                let message =
+                    v1::Message::try_compile(&payer_pubkey, &ixs, recent_blockhash).unwrap();
+                let tx =
+                    VersionedTransaction::try_new(VersionedMessage::V1(message), &[&mint_keypair])
+                        .unwrap();
+                bincode::serialize(&tx).unwrap()
+            })
+            .find(|bytes| {
+                bytes.len() > solana_packet::PACKET_DATA_SIZE
+                    && bytes.len() <= solana_message::v1::MAX_TRANSACTION_SIZE
+            })
+            .expect("a txv1 tx between PACKET_DATA_SIZE and MAX_TRANSACTION_SIZE");
+
+        assert!(serialized.len() > solana_packet::PACKET_DATA_SIZE);
+        assert!(serialized.len() <= solana_message::v1::MAX_TRANSACTION_SIZE);
+
+        let batch = AtomicTxnBatch {
+            seq_id: 1,
+            packets: vec![Packet {
+                data: serialized.into(),
+                meta: None,
+            }],
+            max_schedule_slot: Slot::MAX,
+        };
+
+        let mut stats = BamReceiveAndBufferMetrics::default();
+        let (results, _batch_stats) = run_batch_verify(vec![batch], Slot::MAX, &mut stats);
+
+        assert_eq!(results.len(), 1);
+        // batch_verify calls ed25519_verify with enable_tx_v1=false, so a real V1 tx is
+        // rejected at sigverify and surfaces as DeserializationError. That's fine here.
+        // What matters is the packet reaches sigverify at full length, not truncated at copy.
+        assert!(
+            results[0].is_ok() || matches!(&results[0], Err((Reason::DeserializationError(_), _))),
+            "txv1 tx must reach sigverify at full length, got unexpected result: {:?}",
+            results[0]
+        );
     }
 }
--- a/core/src/banking_stage/transaction_scheduler/bam_receive_and_buffer.rs
+++ b/core/src/banking_stage/transaction_scheduler/bam_receive_and_buffer.rs
@@ -26,8 +26,7 @@ use {
     },
     agave_transaction_view::{
         resolved_transaction_view::ResolvedTransactionView,
-        transaction_version::TransactionVersion,
-        transaction_view::{SanitizedTransactionView, UnsanitizedTransactionView},
+        transaction_version::TransactionVersion, transaction_view::SanitizedTransactionView,
     },
     ahash::HashSet,
     bytes::Bytes,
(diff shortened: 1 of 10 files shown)


core/src/banking_stage/transaction_scheduler/bam_receive_and_buffer.rs
```diff
@@ -727,6 +746,20 @@ impl BamReceiveAndBuffer {
     }
 }
 
+fn bam_packet_data_size_limit(data: &[u8], enable_tx_v1: bool) -> usize {
```
ebin-mathews: Could do something like this
```
fn max_len(data: &[u8], enable_tx_v1: bool) -> usize {
    const V1: u8 = 128 | 1;
    if enable_tx_v1 && data.first() == Some(&V1) {
        MAX_TRANSACTION_SIZE
    } else {
        PACKET_DATA_SIZE
    }
}
```

ebin-mathews: `bam_packet_data_size_limit` parses an `UnsanitizedTransactionView` only to read the version, then `try_new_sanitized` parses the same bytes again. After v1 activation that is the hot path for every packet larger than 1232 bytes.

V1 is identified by the leading version byte `0x81`. A length check is enough here; `verify_transaction_view` already rejects v1 when the feature is off, and the bank enforces size again later.

```rust
fn bam_packet_data_size_limit(data: &[u8], enable_tx_v1: bool) -> usize {
    const V1_VERSION_BYTE: u8 = 128 | 1;
    if enable_tx_v1 && data.first() == Some(&V1_VERSION_BYTE) {
        MAX_TRANSACTION_SIZE
    } else {
        PACKET_DATA_SIZE
    }
}
```

core/src/banking_stage/transaction_scheduler/bam_receive_and_buffer.rs
```diff
@@ -699,10 +715,13 @@ impl BamReceiveAndBuffer {
                 .par_iter_mut()
                 .map(|data| {
                     let data = core::mem::take(data);
+                    if data.len() > bam_packet_data_size_limit(&data, enable_tx_v1) {
```
ebin-mathews: Size and sigverify failures both become `None`, so oversized v1 and disabled-feature v1 look like bad signatures in metrics. Consider counting those drops separately (`tx_v1_disabled` / `oversized`) so activation does not show up as a sigverify cliff.

core/src/banking_stage/transaction_scheduler/bam_receive_and_buffer.rs
```diff
@@ -1135,6 +1205,7 @@ mod tests {
             &thread_pool,
             &mut batches,
             current_slot,
+            false,
```
ebin-mathews: `run_batch_verify` hardcodes `enable_tx_v1: false`, so the new size logic cannot be unit-tested through this helper. Thread the flag through and add cases here instead of only the 30s receive loop:

- feature on: v1 at 1232 / 1233 / 4096 accepted
- feature on: v1 at 4097 dropped
- feature on: legacy/v0 at 1233 dropped
- feature off: large v1 dropped

core/src/banking_stage/transaction_scheduler/bam_receive_and_buffer.rs
```diff
@@ -1065,6 +1100,41 @@ mod tests {
...
+            mint_keypair,
+            ..
+        } = create_slow_genesis_config(u64::MAX);
+        let mut bank = Bank::new_for_tests(&genesis_config);
+        bank.activate_feature(&agave_feature_set::enable_tx_v1::id());
+        let (_bank, bank_forks) = bank.wrap_with_bank_forks_for_tests();
+        (bank_forks, mint_keypair)
+    }
+
+    fn large_v1_transfer_transaction(payer: &Keypair, blockhash: Hash) -> VersionedTransaction {
```
ebin-mathews: `large_v1_transfer_transaction` searches 1..=64 transfers until the wire size is in `(1232, 4096]`. That can land just over 1232 and never hit the 4096 boundary. V1 also caps accounts at 64, so this can fail to find a size.

Pad instruction data to exact sizes (1232, 1233, 4096, 4097), matching the BAM node tests.

core/src/banking_stage/transaction_scheduler/bam_receive_and_buffer.rs
```diff
@@ -1302,6 +1373,46 @@ mod tests {
         }
     }
 
+    #[test]
+    fn test_receive_and_buffer_accepts_large_v1_when_feature_active() {
```
ebin-mathews, line 1528: This only covers “feature on, some v1 larger than 1232, lands.” Please add the feature-off rejection case, or the `enable_tx_v1: false` path can regress without failing CI.

ebin-mathews (commented):
Two follow-ups are outside this diff but still block full v1 support on jito-solana:

- `proto_packet_to_packet` in `core/src/lib.rs` still truncates at 1232 bytes. Relayer and block-engine stages use it, so large v1 packets on those paths will be chopped before sigverify.
- Validator TPU QUIC still defaults to 1232. That does not affect the BAM scheduler stream, but it does affect TPU ingest.

0xEdgar:
1. `proto_packet_to_packet` (core/src/lib.rs): oversized `data` is now discarded, but `meta.size` is still copied unvalidated (`packet.meta_mut().size = meta.size as usize`). A packet whose meta claims more bytes than `data.len()` passes through with a size that points past the backing buffer. Suggest discarding when `meta.size > data.len()`, plus a test.
2. storage-proto decode — storage-proto/src/convert.rs encodes V1+config fine, but decode maps every versioned message back to V0 and drops the config. After activation, V1 transactions round-trip through Bigtable as a different, wrong transaction

0xEdgar:
Two small things on the new storage-proto decode, then I think this PR covers everything:

1. If a stored message somehow has both a `config` and address table lookups, we currently decode it as V1 and silently drop the lookups — so a corrupt record becomes a *different* transaction instead of failing. The encoder never writes that combo, but since the whole point of this ticket is killing silent mutations, could we fall back to V0 (or log loudly) when lookups are non-empty?

2. The round-trip test only covers a fully-populated config. Mind adding the empty case too (config present, all four fields `None`)? That's usually where the default-value paths break.

Neither is big — everything else looks done to me.

esemeniuc (changes requested):
I reviewed the tx-v1 ingest, bundle sigverify, protobuf conversion, local-cluster configuration, and storage-proto paths. The core ingest and storage changes look sound, but I am requesting changes for the ineffective local-cluster feature flag and the missing direct bundle feature-gate coverage.

## Correctness

1. **[P1] `enable_tx_v1 = false` is ineffective.**

   `bam-local-cluster/src/cluster_manager.rs` creates genesis with `FeatureSet::all_enabled()` at line 618, so tx-v1 is already active. `activate_configured_tx_v1_feature()` at line 402 only activates on `true`; `false` leaves the existing feature account in place.

   Explicitly remove the tx-v1 feature account when disabled and test against an all-enabled genesis. If local clusters should always enable tx-v1, remove the flag machinery entirely.

2. **[P2] Bundle tx-v1 feature gating lacks direct coverage.**

   `BundleSigverifyStage` now reads the working bank's feature flag at `core/src/bundle_sigverify_stage.rs:100`, but its tests only exercise legacy packets and shutdown.

   Please add a table-driven test proving:

   - Enabled: a signed V1 packet over 1232 bytes reaches `VerifiedPacketBundle`.
   - Disabled: the same packet is rejected.

3. **[P3] Metrics activity predicates have drifted.**

   `BamReceiveAndBufferMetrics::has_data()` at `core/src/banking_stage/transaction_scheduler/bam_receive_and_buffer.rs:917` checks `total_packets_verified` but omits the new `oversized_packets` and `tx_v1_disabled_packets` counters.

   Give `SigverifyMetrics` one `has_data()` predicate and reuse it from both reporting layers.

## Performance

1. **Replace the linear test transaction-size search with direct sizing.**

   `v1_transaction_with_wire_size()` at `core/src/banking_stage/transaction_scheduler/bam_receive_and_buffer.rs:1158` compiles, signs, and serializes thousands of candidates. The boundary test measured 6.07 seconds by itself.

   V1 instruction lengths are fixed-width, so calculate padding directly:

   `target - (V1 prefix + message.size() + signature_count * SIGNATURE_BYTES)`

   Then compile and sign once.

2. **Drop the working-bank `Arc` before verification.**

   The `working_bank` local at `bam_receive_and_buffer.rs:201` remains alive throughout verification and result parsing despite only providing the slot and feature flag. Extract `(current_slot, enable_tx_v1)` in a short scope. This can also avoid cloning the shared leader bank.

3. **Classify the transaction version once per packet.**

   The verification path checks `is_tx_v1_packet()` and then calls it again through `bam_packet_data_size_limit()` at lines 727-730. Cache `is_v1`, use `solana_message::v1::V1_PREFIX`, and select the size limit with one match.

4. **Separate follow-up: avoid parsing bundles twice.**

   Bundle sigverify parses and sanitizes in `perf/src/sigverify.rs:53`, discards the view, and bundle storage parses it again at `core/src/bundle_stage/bundle_storage.rs:192`. Carrying sanitized views between stages could reduce CPU, but this is invasive and belongs in a separate benchmarked PR.

## Simplicity and LOC

1. **Delete the obsolete 57-line ingest test.**

   `test_txv1_transaction_over_1232_survives_ingest()` at `bam_receive_and_buffer.rs:1906` accepts either success or `DeserializationError`; truncation would produce the permitted error. The stronger boundary and end-to-end tests supersede it.

2. **Remove the size-limit helper and synthetic helper test.**

   Inline the cached V1 classification and limit selection into `batch_verify`. The real signed boundary tests provide stronger coverage. Estimated saving: 15-20 LOC.

3. **Use the canonical protocol constant.**

   Replace the local `128 | 1` definition with `solana_message::v1::V1_PREFIX`.

4. **Use `Result` for packet verification.**

   Replace the custom verified/failure state API with:

   ```rust
   type PacketVerificationResult =
       Result<(SanitizedTransactionView<Bytes>, bool), PacketVerificationError>;
   ```

   This enables `.is_ok()` and eliminates `is_verified()` and some custom state handling.

5. **Derive `Default` for `SigverifyMetrics`.**

   `Histogram::default()` is equivalent to `Histogram::new()`, so the manual implementation at `bam_receive_and_buffer.rs:996` can become `#[derive(Default)]`, saving about 13 LOC.

Approximately 70-90 LOC can be removed without sacrificing meaningful coverage. Removing the local-cluster flag entirely, if feature-off testing is unnecessary, saves another approximately 85 LOC.

Validation performed: all 36 focused tests passed on the PR head and on a clean merge onto current `master`; `git diff --check` passed.

haoranjito:
> I reviewed the tx-v1 ingest, bundle sigverify, protobuf conversion, local-cluster configuration, and storage-proto paths. The core ingest and storage changes look sound, but I am requesting changes for the ineffective local-cluster feature flag and the missing direct bundle feature-gate coverage.
> 
> ## Correctness
> 1. **[P1] `enable_tx_v1 = false` is ineffective.**
>    `bam-local-cluster/src/cluster_manager.rs` creates genesis with `FeatureSet::all_enabled()` at line 618, so tx-v1 is already active. `activate_configured_tx_v1_feature()` at line 402 only activates on `true`; `false` leaves the existing feature account in place.
>    Explicitly remove the tx-v1 feature account when disabled and test against an all-enabled genesis. If local clusters should always enable tx-v1, remove the flag machinery entirely.
> 2. **[P2] Bundle tx-v1 feature gating lacks direct coverage.**
>    `BundleSigverifyStage` now reads the working bank's feature flag at `core/src/bundle_sigverify_stage.rs:100`, but its tests only exercise legacy packets and shutdown.
>    Please add a table-driven test proving:
>    
>    * Enabled: a signed V1 packet over 1232 bytes reaches `VerifiedPacketBundle`.
>    * Disabled: the same packet is rejected.
> 3. **[P3] Metrics activity predicates have drifted.**
>    `BamReceiveAndBufferMetrics::has_data()` at `core/src/banking_stage/transaction_scheduler/bam_receive_and_buffer.rs:917` checks `total_packets_verified` but omits the new `oversized_packets` and `tx_v1_disabled_packets` counters.
>    Give `SigverifyMetrics` one `has_data()` predicate and reuse it from both reporting layers.
> 
> ## Performance
> 1. **Replace the linear test transaction-size search with direct sizing.**
>    `v1_transaction_with_wire_size()` at `core/src/banking_stage/transaction_scheduler/bam_receive_and_buffer.rs:1158` compiles, signs, and serializes thousands of candidates. The boundary test measured 6.07 seconds by itself.
>    V1 instruction lengths are fixed-width, so calculate padding directly:
>    `target - (V1 prefix + message.size() + signature_count * SIGNATURE_BYTES)`
>    Then compile and sign once.
> 2. **Drop the working-bank `Arc` before verification.**
>    The `working_bank` local at `bam_receive_and_buffer.rs:201` remains alive throughout verification and result parsing despite only providing the slot and feature flag. Extract `(current_slot, enable_tx_v1)` in a short scope. This can also avoid cloning the shared leader bank.
> 3. **Classify the transaction version once per packet.**
>    The verification path checks `is_tx_v1_packet()` and then calls it again through `bam_packet_data_size_limit()` at lines 727-730. Cache `is_v1`, use `solana_message::v1::V1_PREFIX`, and select the size limit with one match.
> 4. **Separate follow-up: avoid parsing bundles twice.**
>    Bundle sigverify parses and sanitizes in `perf/src/sigverify.rs:53`, discards the view, and bundle storage parses it again at `core/src/bundle_stage/bundle_storage.rs:192`. Carrying sanitized views between stages could reduce CPU, but this is invasive and belongs in a separate benchmarked PR.
> 
> ## Simplicity and LOC
> 1. **Delete the obsolete 57-line ingest test.**
>    `test_txv1_transaction_over_1232_survives_ingest()` at `bam_receive_and_buffer.rs:1906` accepts either success or `DeserializationError`; truncation would produce the permitted error. The stronger boundary and end-to-end tests supersede it.
> 2. **Remove the size-limit helper and synthetic helper test.**
>    Inline the cached V1 classification and limit selection into `batch_verify`. The real signed boundary tests provide stronger coverage. Estimated saving: 15-20 LOC.
> 3. **Use the canonical protocol constant.**
>    Replace the local `128 | 1` definition with `solana_message::v1::V1_PREFIX`.
> 4. **Use `Result` for packet verification.**
>    Replace the custom verified/failure state API with:
>    ```rust
>    type PacketVerificationResult =
>        Result<(SanitizedTransactionView<Bytes>, bool), PacketVerificationError>;
>    ```
>    
>    
>        
>          
>        
>    
>          
>        
>    
>        
>      
>    This enables `.is_ok()` and eliminates `is_verified()` and some custom state handling.
> 5. **Derive `Default` for `SigverifyMetrics`.**
>    `Histogram::default()` is equivalent to `Histogram::new()`, so the manual implementation at `bam_receive_and_buffer.rs:996` can become `#[derive(Default)]`, saving about 13 LOC.
> 
> Approximately 70-90 LOC can be removed without sacrificing meaningful coverage. Removing the local-cluster flag entirely, if feature-off testing is unnecessary, saves another approximately 85 LOC.
> 
> Validation performed: all 36 focused tests passed on the PR head and on a clean merge onto current `master`; `git diff --check` passed.

Done. 

Didn't address double-parse. it is out of scope of this pr. leave it for future pr.

esemeniuc (approved):

PR #1553, opened 2026-08-13
Fix Block Engine autoconfig candidate failover

esemeniuc:
## Summary

- build a fresh endpoint for every ranked Block Engine candidate
- publish the matching Shredstream receiver only after authentication and stream setup succeeds
- clear stale Shredstream metadata when a successful candidate has no resolvable receiver
- report the actual global endpoint in direct-mode disconnect metrics

## Issue
The autoconfig loop retained a mutable endpoint initialized from the configured URL and rebuilt it only when a candidate URL differed from that configuration. If a different candidate was attempted first and the configured candidate appeared later, the later iteration reused the previous candidate's endpoint while carrying the configured candidate's Shredstream metadata. The loop also published each candidate's Shredstream receiver before confirming that its authenticated streams could be established.

## New Behaviour

Fallback attempts now connect to the candidate being iterated, and the active Shredstream destination stays paired with the Block Engine session that completed setup.

### DIFF
--- a/core/src/proxy/block_engine_stage.rs
+++ b/core/src/proxy/block_engine_stage.rs
@@ -376,16 +376,9 @@ impl BlockEngineStage {
 
         // try connecting to best block engine
         let endpoint_count = candidates.len();
-        for (block_engine_url, (maybe_shredstream_socket, latency_us)) in candidates
-            .into_iter()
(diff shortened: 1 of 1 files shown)


haoranjito (approved):
lgtm

PR #1554, opened 2026-08-13
rpc: add filtered fan-out for bank notifications

hash-envy:
#### Problem
- The upcoming changes to support a `tip-router-snapshot-mode` in jito-solana require access to `BankNotifications` 

#### Summary of Changes
- Allows for multiple subscribers of BankNotifications
- Adds simple filtering for notifications


rpc/src/optimistically_confirmed_bank_tracker.rs
```diff
@@ -81,11 +81,121 @@ pub type BankNotificationWithDependencyWork = (
...
+
+    pub fn should_forward(&self, notification: &BankNotification) -> bool {
+        self.filter
+            .as_ref()
+            .is_none_or(|filter| filter.should_forward(notification))
+    }
+
+    /// Forward `notification` to this subscriber, returning whether it was delivered.
+    pub fn send_notification(&self, notification: BankNotificationWithDependencyWork) -> bool {
+        match self.tx.send(notification) {
```
morgyang: **HIGH — a subscriber with a bounded channel will block consensus threads here.**

`Sender::send` on a bounded crossbeam channel blocks until capacity frees. Every producer that reaches this call is consensus-critical:

| Producer | Site | Notification |
|---|---|---|
| Replay thread | `core/src/replay_stage.rs:4139` | `Frozen` |
| Gossip vote listener | `core/src/cluster_info_vote_listener.rs:759` | `OptimisticallyConfirmed` |
| Rooting path | `votor/src/root_utils.rs:165,177` | `NewRootBank`, `NewRootedChain` |
| Votor `set_root` | `votor/src/root_utils.rs:85` | `OptimisticallyConfirmed` |

`new()` (L102) and `new_with_filter()` (L107) accept any `Sender`, so nothing prevents a subscriber from passing `bounded(N)` — and reaching for a bounded channel to get backpressure is the *normal* engineering instinct. That's what makes this a trap rather than an edge case: the responsible-looking choice is the dangerous one.

The filter offers no protection here, because the hazard is on the notifications the subscriber actually *wants*.

On blast radius — I checked, and no `bank_forks` lock is held across any of these sends. In `check_and_handle_new_root` the write lock is taken and released inside `set_bank_forks_root` (`votor/src/root_utils.rs:214`) before the sends at `:160-183`, and the other sites take only momentary read locks. So this is a thread-stall hazard, not a lock-hold one.

**Suggested fix:** `try_send` with a drop counter, or have `new`/`new_with_filter` construct the channel internally and hand back the `Receiver`, so unboundedness is enforced by construction rather than by convention.

rpc/src/optimistically_confirmed_bank_tracker.rs
```diff
@@ -81,11 +81,121 @@ pub type BankNotificationWithDependencyWork = (
...
 
 pub type BankNotificationReceiver = Receiver<BankNotificationWithDependencyWork>;
-pub type BankNotificationSender = Sender<BankNotificationWithDependencyWork>;
+
+/// Decides whether a subscriber should receive a given bank notification.
+///
+/// This method runs synchronously on replay, gossip, and root-processing producer threads.
+/// Implementations must return quickly and must not perform blocking I/O, wait on contended locks,
+/// or do expensive work. A slow or blocking filter delays bank-notification production for every
+/// subscriber.
```
morgyang, line 90: Follow-on to the HIGH below: this doc correctly warns that the filter must not block — right instinct, but applied to the less likely hazard. The channel is the bigger one and isn't mentioned here. Worth adding a line requiring the subscriber's channel to be unbounded, or making that structurally impossible via the constructor.

rpc/src/optimistically_confirmed_bank_tracker.rs
```diff
@@ -81,11 +81,121 @@ pub type BankNotificationWithDependencyWork = (
...
+            // channel was never touched.
+            if !sender.should_forward(&notification.0) {
+                connected_subscriber_count += 1;
+                continue;
+            }
+
+            if sender.send_notification(notification.clone()) {
+                connected_subscriber_count += 1;
+            }
+        }
```
morgyang, line 201: **MED — retention here is plausible but unobservable.**

`notification.clone()` on L182 hands a live `Arc<Bank>` to each subscriber's queue (`Frozen` and `NewRootBank` both carry one). An unfiltered subscriber that stalls accumulates one per slot, ~2.5/sec.

The interaction with pruning is worth calling out: a retained `Arc` does **not** stop `bank_forks` from dropping the entry — `get_non_rooted` (`runtime/src/bank_forks.rs:709-723`) removes it regardless — but the memory isn't reclaimed until the last `Arc` goes. So pruning appears to succeed while memory stays resident, which makes this hard to attribute after the fact.

I'd have filed this as HIGH, but the filter defuses most of it: a subscriber correctly filtered to epoch-boundary roots receives roughly one notification per epoch, so even a multi-hour stall costs one extra bank. What's genuinely missing from this PR isn't a bound, it's **observability** — no queue-depth metric, no lag counter, no drop policy.

Two things would push it back to HIGH: a first subscriber that ships with `new()` instead of `new_with_filter()`, or a filter broad enough to admit per-slot `Frozen`.

**Suggested fix:** a per-subscriber datapoint for send count and queue depth, landed before anything actually subscribes.

rpc/src/optimistically_confirmed_bank_tracker.rs
```diff
@@ -81,11 +81,121 @@ pub type BankNotificationWithDependencyWork = (
...
+    ) -> Result<(), BankNotificationBroadcastError> {
+        let mut connected_subscriber_count = 0;
+
+        for sender in self.subscriber_senders.iter() {
+            // Filter before cloning, so a rejected notification never clones the bank it carries.
+            // A subscriber that filters the notification out is still assumed connected, since its
+            // channel was never touched.
+            if !sender.should_forward(&notification.0) {
+                connected_subscriber_count += 1;
+                continue;
```
morgyang: **MED — a filtered-out subscriber is counted as connected.**

Its channel was never touched, so liveness here is unverified, not observed. Combined with the `== 0` check below, a broadcaster whose only *unfiltered* subscriber has died will keep reporting `Ok` indefinitely.

rpc/src/optimistically_confirmed_bank_tracker.rs
```diff
@@ -81,11 +81,121 @@ pub type BankNotificationWithDependencyWork = (
...
+
+            if sender.send_notification(notification.clone()) {
+                connected_subscriber_count += 1;
+            }
+        }
+
+        if connected_subscriber_count == 0 {
+            warn!("bank notification broadcast failed: no connected subscribers");
+            return Err(BankNotificationBroadcastError);
+        }
```
morgyang, line 206: **MED — returning `Ok` when any subscriber survives hides the one that matters.**

Every call site does `unwrap_or_else(|err| warn!(...))` — `core/src/replay_stage.rs:4153`, `core/src/cluster_info_vote_listener.rs:774`, `votor/src/root_utils.rs:168,181`. So if `OptimisticallyConfirmedBankTracker` dies while an auxiliary subscriber lives, all of them see `Ok`, and the only trace is a single `warn!` buried in `send_notification`.

That tracker drives RPC commitment and `PrioritizationFeeCache` cleanup. Its death should be loud, not folded into an aggregate bool.

**Suggested fix:** give each `BankNotificationSender` a name/label and report per-subscriber outcomes, with a counter rather than a single boolean.

rpc/src/optimistically_confirmed_bank_tracker.rs
```diff
@@ -81,11 +81,121 @@ pub type BankNotificationWithDependencyWork = (
...
+
+    /// Forward `notification` to this subscriber, returning whether it was delivered.
+    pub fn send_notification(&self, notification: BankNotificationWithDependencyWork) -> bool {
+        match self.tx.send(notification) {
+            Ok(()) => true,
+            Err(SendError(notification)) => {
+                warn!(
+                    "bank notification subscriber disconnected, dropping {:?}",
+                    notification.0
+                );
```
morgyang: **MED — unthrottled, for as long as the subscriber stays dead.**

No rate limit and no "already reported" state, so a disconnected subscriber emits this on *every* notification. At ~10–12 notifications/sec across all producers that's roughly 1M lines/day, indefinitely.

Shutdown is the benign case — a second or two of noise while producers wind down (three lines per notification once you count L188 and the call site). The case that matters is a subscriber thread that dies during normal operation: nothing restarts it and nothing suppresses the log, so a week-long validator run buries real errors under tens of millions of identical warnings.

**Suggested fix:** latch per subscriber — report the first disconnect, then increment a counter.

rpc/src/optimistically_confirmed_bank_tracker.rs
```diff
@@ -81,11 +81,121 @@ pub type BankNotificationWithDependencyWork = (
...
+            subscriber_senders: subscriber_senders.into(),
+        }
+    }
+
+    /// Broadcasts a notification to each matching subscriber.
+    ///
+    /// Concurrent calls are not serialized across subscribers. Each subscriber's channel
+    /// preserves its own send order, but different subscribers may observe concurrent
+    /// notifications in different relative orders because their fan-out loops can interleave.
+    /// Consumers must not rely on a common global ordering across subscriber channels.
```
morgyang, line 188: **MED — the `dependency_work` contract isn't stated anywhere in the new API.**

This doc covers cross-subscriber ordering, but not the `Option<u64>` riding along in `BankNotificationWithDependencyWork`. That value is a gate: `process_notification` calls `wait_for_dependency` before acting, so the tracker never observes state ahead of `TransactionStatusService` (which marks work processed at `rpc/src/transaction_status_service.rs:265`).

The broadcaster now hands the same id to every subscriber with no indication of what it's for. A subscriber that ignores it will act on banks whose transaction statuses haven't been written yet.

Worth adding to the trait doc on L85-93 as well: subscribers at this layer see **raw** notifications, upstream of the tracker's dedup (`highest_confirmed_slot`), defer (`pending_optimistically_confirmed_banks`), and hash-mismatch handling. Duplicate and out-of-order OC notifications for the same slot are normal here and have to be handled downstream.

rpc/src/optimistically_confirmed_bank_tracker.rs
```diff
@@ -81,11 +81,121 @@ pub type BankNotificationWithDependencyWork = (
...
+
+#[derive(Clone, Debug)]
+pub struct BankNotificationBroadcaster {
+    subscriber_senders: Arc<[BankNotificationSender]>,
+}
+
+#[derive(Debug, PartialEq, Eq)]
+pub struct BankNotificationBroadcastError;
+
+impl BankNotificationBroadcaster {
```
morgyang, line 176: **MED — no test coverage on ~120 lines of new public API.**

Cheap cases worth adding:

- filter accepts → delivered; filter rejects → not delivered
- all subscribers disconnected → `Err(BankNotificationBroadcastError)`
- one of two disconnected → `Ok`
- a filtered-out subscriber never clones the notification (the property the comment on L174-176 asserts)

morgyang (commented):
Reviewed the fan-out mechanism and traced it against the current call sites. **The change is necessary and the shape is right** — the `BankNotification` channel is single-consumer, and a snapshot subscriber needs `Arc<Bank>` at root, which no existing subscription point provides.

Worth putting the "why not `slot_notification_subscribers`" answer into the PR description, since it's the first question a reviewer will ask. Three independent reasons, any one of which is disqualifying:

1. **The bank is gone.** `SlotNotification` is `(slot, parent, bank_id)`. Looking the bank up afterwards via `bank_forks.get(slot)` races pruning — which is exactly what holding the `Arc` prevents.
2. **Delivery is conditional.** `SlotNotification::Frozen` only fires inside `if let Some(parent) = bank.parent()`; the OC path defers into `pending_optimistically_confirmed_banks` on hash mismatch and drops outright once the slot is already rooted. An epoch-boundary snapshotter can't tolerate a silently skipped root.
3. **It sits downstream of a blocking wait.** `process_notification` calls `wait_for_dependency` before dispatching anything, so every slot-notification subscriber inherits a stall on `TransactionStatusService`.

One framing note on the filter: it currently reads as a throughput optimization, but total volume across all producers is only ~10–12 msg/sec — OC notifications are edge-triggered once per `(slot, hash)` (`core/src/consensus/vote_stake_tracker.rs:29-31`), not once per vote. Its real value is `Arc<Bank>` retention, which is what the comment on L174-176 is getting at. Stating that explicitly is what will tell the next subscriber author to supply a filter rather than defaulting to `new()`.

**Findings below: 1 HIGH, 5 MED.** Only the HIGH is filter-proof. LOW-severity nits omitted — happy to add them if useful.

Unqualified line references are post-merge positions in this file; cross-file references name the file.

rpc/src/optimistically_confirmed_bank_tracker.rs
hash-envy: This was considered. I had to balance correctness, with maintainability here. Since this structure is not upstream, I do not anticipate anyone adding any additional subscribers. The two that exist rn are both `unbounded` and if upstream changes this, they would have to account for this issue as well. 

The alternative is, as you suggest, switching to a `try_send` which increases the surface area of code we will have to reason about if upstream changes affect the original implementation. 

Though forcing correctness on construction might be less of a change. Good suggestion, ill look into it.

rpc/src/optimistically_confirmed_bank_tracker.rs
hash-envy: The good news is that there will only be 2 paths where this can occur: 
- The existing consumer of the notifications, which does not take this issue into account
- The tip-router-service. If it does block, yes that would be an issue, but it would at least be confined to people running with `--tip-router-service true` (currently 8 operators only), and these operators are explicitly NOT able to also have `--is-voting true` set. 

`send_notification` is non-blocking, so the original notification will be dropped immediately after sending to all the subscribers. 

I'll look into adding better observability for it.

rpc/src/optimistically_confirmed_bank_tracker.rs
hash-envy: It's schrodinger's channel. If we don't need to send something to it, do we need to observe and warn if it has been dropped? I don't think so.

rpc/src/optimistically_confirmed_bank_tracker.rs
hash-envy: Yeah. I'll fix this

rpc/src/optimistically_confirmed_bank_tracker.rs
hash-envy: Added correctness via construction

rpc/src/optimistically_confirmed_bank_tracker.rs
hash-envy: Forcing it + added comment why

esemeniuc (dismissed):
some feedback to address
```
 Findings, highest priority first:

  1. Medium — the filtered-channel contract is bypassable.
     channel_with_filter() promises filtered delivery (https://github.com/jito-foundation/jito-solana/blob/47e80668b923ae44f9a944fd93055be3f28aa8e2/rpc/src/optimistically_confirmed_bank_tracker.rs#L106-L110), but public
     send_notification() (https://github.com/jito-foundation/jito-solana/blob/47e80668b923ae44f9a944fd93055be3f28aa8e2/rpc/src/optimistically_confirmed_bank_tracker.rs#L128-L130) sends directly without applying the filter. An
     external caller can therefore create a filtered sender and deliver rejected notifications. Both should_forward and send_notification are only used inside this module; making them private would restore the invariant.

  2. Medium — stalled subscribers can retain unbounded bank state.
     Every subscriber channel is intentionally unbounded (https://github.com/jito-foundation/jito-solana/blob/47e80668b923ae44f9a944fd93055be3f28aa8e2/rpc/src/optimistically_confirmed_bank_tracker.rs#L113-L119), and each matching
     send clones the notification (https://github.com/jito-foundation/jito-solana/blob/47e80668b923ae44f9a944fd93055be3f28aa8e2/rpc/src/optimistically_confirmed_bank_tracker.rs#L178-L188). Frozen and NewRootBank carry Arc<Bank>, so
     a stalled unfiltered subscriber can retain banks indefinitely with no queue-depth or lag metric. This is latent until another subscriber is wired in, but that is the purpose of this API. This concern remains unresolved
     (https://github.com/jito-foundation/jito-solana/pull/1554#discussion_r3799658574).

  3. Medium — subscriber death produces ambiguous success and unlimited warning spam.
     A dead subscriber is never removed or latched, so warn! runs for every subsequent matching notification
     (https://github.com/jito-foundation/jito-solana/blob/47e80668b923ae44f9a944fd93055be3f28aa8e2/rpc/src/optimistically_confirmed_bank_tracker.rs#L128-L137). Meanwhile, the broadcaster returns Ok if any other subscriber
     survives—and even counts a filtered-out subscriber as connected. Thus the essential RPC tracker can die while callers continue receiving Ok; the warning does not identify which subscriber failed. The author previously said this
     would be fixed, but the thread remains unresolved (https://github.com/jito-foundation/jito-solana/pull/1554#discussion_r3799658591).

  The core fan-out loop is otherwise sound:

  - Bounded channels are now impossible through the constructors, avoiding producer-thread blocking.
  - Filtering happens before cloning.
  - The documented cross-subscriber ordering limitation is accurate.
  - With the currently configured single unfiltered subscriber, normal delivery remains equivalent to the original implementation.
```

## Simplification and performance improvements

The fan-out can be made smaller and faster without adding helper-only abstractions.

### Recommended design

1. Replace the public `should_forward()` and `send_notification()` methods with one private `forward(&BankNotificationWithDependencyWork)` operation. It should check disconnection, apply the filter, clone only accepted notifications, and send. This restores the filtered-channel invariant and keeps filter-before-clone behavior in one place.

2. Remove `BankNotificationBroadcastError`, `connected_subscriber_count`, the filtered-out-as-connected branch, and the final success/error branch. No producer changes control flow based on the aggregate result; callers only log it, while partial delivery makes the result ambiguous. Make broadcaster `send()` return `()` and remove the producer-side `.unwrap_or_else(...)` calls.

3. Add an `AtomicBool` disconnect latch to each subscriber. After the first failed send, log once and skip future filtering, cloning, send attempts, and warnings. A disconnected crossbeam receiver cannot reconnect, so repeated attempts are pure overhead.

4. Remove the unused manual `Debug` implementation for `BankNotificationSender` and remove `Debug` from `BankNotificationBroadcaster`. Nothing formats either type.

This should remove roughly 25–35 lines even after adding disconnect latching. Keep the broadcaster as the meaningful fan-out abstraction; folding it into the config or adding forwarding constructors would mostly move code around.

### Performance requirements for auxiliary subscribers

- Keep `OptimisticallyConfirmedBankTracker` first in the subscriber slice so its notification is queued before auxiliary filters run.
- The tip-router filter should reject `Frozen`, optimistic-confirmation, and ordinary-root traffic before cloning, and accept only the epoch-boundary root events it requires.
- Filters must remain lock-free and inexpensive because they execute synchronously on replay, gossip, and root-processing producer threads.
- Drain the receiver immediately and hand expensive snapshot work to a separate worker. Do not retain queued `Arc<Bank>` values longer than necessary.
- Sample queue depth and receiver lag periodically; do not emit metrics for every notification.

### Conditional optimizations

- `NewRootedChain(Vec<(Slot, BankId)>, Slot)` is the only notification whose clone performs a deep allocation/copy. If multiple subscribers actually consume it, sort once at production and use an immutable `Arc<[(Slot, BankId)]>` payload.
- If auxiliary notifications may be coalesced, use a bounded nonblocking or latest-value queue. Never use blocking `send()` on a bounded channel from these producer threads.

Avoid optimizing dynamic filter dispatch, `Arc<[BankNotificationSender]>`, or the final subscriber clone. Also avoid iterator constructs such as `.any()`, which would short-circuit and break fan-out. The direct loop plus one private `forward()` operation is the simplest useful implementation.

rpc/src/optimistically_confirmed_bank_tracker.rs
hash-envy: Added

rpc/src/optimistically_confirmed_bank_tracker.rs
hash-envy: Added atomic bool (and subscriber labels)

rpc/src/optimistically_confirmed_bank_tracker.rs
hash-envy: From what i can tell, all current producers just log or drop the error entirely. And producers _should_ keep sending notifications until all the channels are dead, since the other channels rely on them. They should not error if one is dead.

rpc/src/optimistically_confirmed_bank_tracker.rs
hash-envy: It's an unbounded channel, so I'm not too sure there is a great way of adding built in observability to it. We could technically add a `send_count` and a `recevied_count` on each receiver and set up an alert on the difference, but I really think that especially since there is only really 1 service that can cause this issue (tip-router), we are better off just being extra cautious about failure cases for tip-router (there currently exist timeouts to ensure we don't hold banks too long).

esemeniuc:
## Consolidated review report
No new high-severity delivery bug was found, and the core fan-out loop is sound, but several medium and low issues remain on `98238d86`.

### Findings

1. **Medium — disconnect logging is still unbounded.**

   The new atomic latch suppresses only the warning inside [`BankNotificationSender::forward()`](https://github.com/jito-foundation/jito-solana/blob/98238d86c4b83d4e173c6cf54a92a54e43c1aff2/rpc/src/optimistically_confirmed_bank_tracker.rs#L136-L164). Once all subscribers disconnect, [`BankNotificationBroadcaster::send()`](https://github.com/jito-foundation/jito-solana/blob/98238d86c4b83d4e173c6cf54a92a54e43c1aff2/rpc/src/optimistically_confirmed_bank_tracker.rs#L189-L208) still warns and returns `Err` on every notification, and most producers [log the error again](https://github.com/jito-foundation/jito-solana/blob/98238d86c4b83d4e173c6cf54a92a54e43c1aff2/core/src/replay_stage.rs#L4147-L4153).

   With the currently configured single subscriber, this still produces one or two warnings per event indefinitely. The comment claiming that the latch avoids all subsequent “logging work” is therefore incorrect.

   The simplest fix is to make broadcaster `send()` return `()`, remove `BankNotificationBroadcastError` and `connected_subscriber_count`, and use the labeled, one-shot warning plus a disconnect metric as the sole failure signal.

2. **Medium — the aggregate result has misleading semantics.**

   A filtered-out subscriber counts as connected, so `send()` can return `Ok` without delivering the notification. It also returns `Ok` when the essential optimistically-confirmed-bank tracker dies but an auxiliary subscriber remains alive.

   Subscriber labels make the initial failure identifiable, but the aggregate `Result` still provides no reliable delivery or critical-service-liveness guarantee. No producer changes control flow based on it; callers only log or discard it. Removing the result would both simplify the code and fix the repeated outer warnings.

3. **Medium — stalled subscribers can retain unbounded bank state without observability.**

   Channels are correctly forced to be unbounded, but every accepted notification is cloned into the subscriber queue. `Frozen` and `NewRootBank` carry `Arc<Bank>`, so a stalled subscriber can retain banks indefinitely even after `BankForks` pruning appears to succeed.

   There is still no queue-depth, lag, or disconnect metric. Crossbeam exposes queue depth directly through `Sender::len()`, so it can be sampled periodically per subscriber without maintaining separate sent/received counters.

   This PR still configures only the original tracker, so the additional retention risk becomes active when an auxiliary subscriber is wired in.

4. **Medium — the dependency and raw-notification contracts remain undocumented.**

   Despite the review reply saying this was added, the public API still has only [`Option<u64> // dependency work id`](https://github.com/jito-foundation/jito-solana/blob/98238d86c4b83d4e173c6cf54a92a54e43c1aff2/rpc/src/optimistically_confirmed_bank_tracker.rs#L78-L93).

   Subscribers need to know that:

   - The work ID may need to be awaited before reading transaction-status data.
   - Notifications are delivered before the existing tracker’s deduplication, deferral, and hash-mismatch processing.
   - Duplicate or out-of-order optimistic-confirmation events must be handled downstream.

5. **Low — `NewRootedChain` now incurs an unnecessary deep clone.**

   [`forward()` clones every accepted notification](https://github.com/jito-foundation/jito-solana/blob/98238d86c4b83d4e173c6cf54a92a54e43c1aff2/rpc/src/optimistically_confirmed_bank_tracker.rs#L151), including when only one subscriber exists. `Frozen` and `NewRootBank` only increment an `Arc` reference count, but cloning `NewRootedChain(Vec<...>)` allocates and copies the vector. Before this PR, that vector was moved directly into the channel.

   This affects rooted-chain notifications when Geyser is enabled. If multiple consumers need them, sort once at production and use an immutable `Arc<[(Slot, BankId)]>` payload. Otherwise, avoid adding complex last-subscriber machinery solely for this optimization.

6. **Low — the filter contract should prohibit panics explicitly.**

   The documentation correctly prohibits slow or blocking filters, but a panic propagates directly through replay, gossip, or root-processing threads. State that filter implementations must not panic. Adding `catch_unwind` is likely unnecessary.

7. **Low — multi-subscriber behavior remains untested.**

   The only new test [checks that the channel is unbounded](https://github.com/jito-foundation/jito-solana/blob/98238d86c4b83d4e173c6cf54a92a54e43c1aff2/rpc/src/optimistically_confirmed_bank_tracker.rs#L622-L626). Missing coverage includes:

   - Filter acceptance and rejection.
   - Delivery to multiple matching subscribers.
   - Partial and complete disconnection.
   - Disconnect latching.
   - Filter-before-clone behavior.
   - Aggregate result behavior, if the result is retained.

### Confirmed resolved

- Bounded channels can no longer block producer threads.
- Filtered delivery cannot be bypassed through a public direct-send method.
- Filtering occurs before cloning.
- A disconnected sender skips subsequent filter, clone, and send attempts.
- The per-subscriber disconnect warning is emitted once and includes a label.
- The unnecessary manual `Debug` implementation was removed.
- The documented cross-subscriber ordering limitation is accurate.

No additional event-loss, synchronization, or atomic-ordering defect was found in the core fan-out implementation.

hash-envy:
@esemeniuc  Below I address the latest findings 1 by 1, though there are two themes I want to outline first:

A) I don't want to increase the surface area of code that differs from upstream, in order to patch issues that existed prior to these changes

B) Most of these potential edge cases will only effect nodes running the tip-router extension (of which there are not many). Running the tip router service will be mutually exclusive with running a voting node

1) "Medium — disconnect logging is still unbounded."
 - Correct, but this is the behavior upstream as well. Upstream it is a raw crossbeam channel that errors and emits a warning at each producer site. None of the producers handle the error except for logging it.
 - Technically we've added an additional `warn` that will fire when all subscribers are disconnected. I'm not convinced this is worth handling rn -- I think it's probably helpful to know how many times producers fire after the connection dies. 
2) "Medium — the aggregate result has misleading semantics" but this is followed by "No producer changes control flow based on it; callers only log or discard it."
- I agree with the second bit -- and i do not want to increase the surface area diverged from upstream

3) Medium — stalled subscribers can retain unbounded bank state without observability.
- Yes. This is true. I am _very_ careful to drop Bank quickly in the tip router branch, and we don't block on incoming messages. Additionally, Banks are typically held by the system until rooted (my understanding is that this is ~15 sec rn), and our TipRouter Filter will only forward epoch boundary Banks (so less than 0.01% of Banks are held). 
- I added a metric for subscriber channel depth

4) "Medium — the dependency and raw-notification contracts remain undocumented"
- My bad -- i hadn't pushed the commit yet

5) "Low — NewRootedChain now incurs an unnecessary deep clone"
- It should never have more than a few slots in it. This is not a large vec. 

6) "Low — the filter contract should prohibit panics explicitly."
Ill add a comment 

7) "Low — multi-subscriber behavior remains untested." 
- Added

core/src/validator.rs
```diff
@@ -1415,7 +1416,7 @@ impl Validator {
                     dependency_tracker.clone(),
                 ));
             let bank_notification_sender_config = Some(BankNotificationSenderConfig {
-                sender: bank_notification_sender,
+                sender: BankNotificationBroadcaster::new(vec![bank_notification_sender]),
```
buffalu, line 1419: couldn't we just turn this from a single sender to vector of senders? then change the caller that calls bank_notification_sender.send to:
```rs
for sender in &senders {
     sender.try_send(bank.clone())
     // or sender.send(bank.clone())
}
```

think it would make this code much simpler, assuming the readers can read fast enough

rpc/src/optimistically_confirmed_bank_tracker.rs
```diff
@@ -81,11 +82,158 @@ pub type BankNotificationWithDependencyWork = (
...
+                tx,
+                filter,
+                disconnected: AtomicBool::new(false),
+            },
+            rx,
+        )
+    }
+
+    /// Forward `notification` if accepted by this subscriber's filter, returning whether the
+    /// subscriber is assumed connected.
```
0xEdgar, line 150: what happens if we end up with a slow filter?

rpc/src/optimistically_confirmed_bank_tracker.rs
```diff
@@ -464,9 +612,31 @@ mod tests {
         crossbeam_channel::bounded,
         solana_ledger::genesis_utils::{GenesisConfigInfo, create_genesis_config},
         solana_runtime::{bank::SlotLeader, commitment::BlockCommitmentCache, dependency_tracker},
-        std::sync::atomic::AtomicU64,
+        std::sync::atomic::{AtomicU64, AtomicUsize},
     };
 
+    struct FixedBankNotificationFilter(bool);
```
0xEdgar, line 618: my understanding is that there's only going to be one anticipated consumer and one predicate. with that in mind, wonder if we can get away with less API:

```rust
filter: Option<
    Box<dyn Fn(&BankNotification) -> bool + Send + Sync + 'static>
>
```

rpc/src/optimistically_confirmed_bank_tracker.rs
0xEdgar: i assume we're gonna have something like `is_epoch_boundary` ; is that an expensive check?

rpc/src/optimistically_confirmed_bank_tracker.rs
```diff
@@ -81,11 +82,158 @@ pub type BankNotificationWithDependencyWork = (
...
+
+        if !self
+            .filter
+            .as_ref()
+            .is_none_or(|filter| filter.should_forward(&notification.0))
+        {
+            return true;
+        }
+
+        match self.tx.send(notification.clone()) {
```
0xEdgar, line 166: this clone looks to be okay, but wanna make sure.
`NewRootedChain` seems to be a deep clone, but probably cheep.

rpc/src/optimistically_confirmed_bank_tracker.rs
```diff
@@ -81,11 +82,158 @@ pub type BankNotificationWithDependencyWork = (
...
+        }
+
+        Ok(())
+    }
+}
 
 #[derive(Clone)]
 pub struct BankNotificationSenderConfig {
-    pub sender: BankNotificationSender,
+    pub sender: BankNotificationBroadcaster,
```
0xEdgar, line 236: Would it be simpler for this config to own senders: `Vec<BankNotificationSender>` directly, initialized with `senders: vec![bank_notification_sender]`?

- benefit is a more explicit data model with one fewer wrapper type
- tradeoff: extra clones. `BankNotificationSenderConfig` is cloned into several producer contexts, but i think that's fine.

0xEdgar (commented):
some comments. overall looks fine

rpc/src/optimistically_confirmed_bank_tracker.rs
hash-envy: Yes but then all the producers have to loop through the Vec, and each producer has to keep track of subscriptions dropping. 

The explicit benefit of this model is that it doesn't touch the existing upstream producer api. I wanted to change as little upstream code as possible to make fork maintenance easier

core/src/validator.rs
hash-envy: I was trying to reduce the amount of interfaces touched (and make merging upstream changes in easier). 

If we do it this way, then each producer has to loop, ensure that bank_notification_sender is _first_ in the Vec, and handle any dropped subscriptions. 

Having an explicit `BankNotificationBroadcaster` struct reduces duplication. Though yes, I do think its simple enough that it may not be that big of a deal.

rpc/src/optimistically_confirmed_bank_tracker.rs
hash-envy: If we have a slow filter, it is bad. We warn about it in comments. 

That being said, no one upstream will ever be using this primitive, and I doubt we will have any other consumers of it except for TipRouter. And even if TipRouter has a slow filter, it will only affect TipRouter nodes. 

This is what i have for epoch_boundary. It's cheap. 
```rust
impl BankNotificationFilter for TipRouterEpochBoundaryFilter {
    fn should_forward(&self, notification: &BankNotification) -> bool {
        match notification {
            BankNotification::Frozen(bank) => {
                // A boundary is a bank whose epoch is greater than its parent's. Deriving the
                // parent's epoch from `parent_slot` rather than `parent()` keeps this free of the
                // lock and `Arc` clone that `Bank::parent` costs, and still holds when the first
                // slots of an epoch are skipped.
                bank.epoch() > bank.epoch_schedule().get_epoch(bank.parent_slot())
            }
            BankNotification::NewRootedChain(..) => true,
            BankNotification::OptimisticallyConfirmed(..) | BankNotification::NewRootBank(_) => {
                false
            }
        }
    }
```

rpc/src/optimistically_confirmed_bank_tracker.rs
hash-envy: I personally think trait + struct is a tad cleaner / easier to read, but i have no qualms with this approach if you prefer it.

rpc/src/optimistically_confirmed_bank_tracker.rs
hash-envy:
```
 NewRootedChain(Vec<(Slot, BankId)>, Slot),
```

Slot and BankId are u64 typedef/newtypes. The Vec should never be more than a few elements. And even if it is, it will never be faster than slot latency (200ms minimum ), so it should not be a heavy clone that blocks anything.

0xEdgar (approved):

PR #1561, opened 2026-08-14
Prevent stale tip-crank configuration across BAM/Block Engine transitions

esemeniuc:
#### Problem

BundleStage and BankingStage used separate block-builder fee snapshots with different update paths. After BAM connected, BundleStage could retain a stale Block Engine snapshot while BankingStage used BAM's current snapshot. Their independent per-slot guards allowed both stages to submit conflicting block-builder cranks in the same leader slot.

This was a reachable correctness issue that could temporarily publish a stale builder and commission, creating potential tip attribution and accounting impact. The investigation identified the code path that allowed the issue; it did not establish a confirmed production occurrence.

#### Summary of Changes

- Share one initially-empty `ArcSwapOption<BlockBuilderFeeInfo>`
- Validate the complete pubkey/commission pair before publishing.
- Add regressions for source ownership, replaced BAM snapshots, and stale fallback state.

### DIFF
--- a/core/src/bundle_stage.rs
+++ b/core/src/bundle_stage.rs
@@ -830,7 +830,7 @@ impl BundleStage {
     ) -> BundleExecutionResult<()> {
         let block_builder_fee_info = block_builder_fee_info.load();
         if block_builder_fee_info.block_builder == Pubkey::default() {
-            return Err(BundleExecutionError::TipError);
+            return Err(BundleExecutionError::ErrorRetryable);
         }
         let crank_tip_program_transactions = tip_manager
             .get_tip_programs_crank_bundle(bank, keypair, &block_builder_fee_info)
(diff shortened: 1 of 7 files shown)


core/src/bundle_stage.rs
```diff
@@ -829,6 +829,9 @@ impl BundleStage {
         keypair: &Keypair,
     ) -> BundleExecutionResult<()> {
         let block_builder_fee_info = block_builder_fee_info.load();
+        if block_builder_fee_info.block_builder == Pubkey::default() {
+            return Err(BundleExecutionError::TipError);
```
0xEdgar: This returns through the catch-all arm in `consume_bundles` (core/src/bundle_stage.rs:643-650), which sets `*last_tip_update_slot = bank.slot()`. The default sentinel is transient though — if the Block Engine publishes the fee info later in this same slot (first leader slot after a BAM→BE cutover, or on startup), subsequent bundles in the slot skip `handle_tip_programs` entirely and execute against the previous leader's on-chain tip config — the stale-attribution case this PR is closing.

`ErrorRetryable` takes the arm that returns without marking the slot, so the crank retries on the next 10ms poll once the snapshot is published:

```suggestion
            return Err(BundleExecutionError::ErrorRetryable);
```

(The retryable arm logs `error!` every 10ms poll — may be worth debouncing/downgrading that log for this case.)

esemeniuc: done

0xEdgar (approved):

PR #1572, opened 2026-08-18
Remove dead Jito artifacts

esemeniuc:
## What changed

- remove the orphaned `transaction-metrics-tracker` source, which has no manifest or references
- remove the undeclared `bundle_stage_leader_metrics` module, which never compiled or emitted metrics
- remove four historical tip-program binaries that are not included or fetched anywhere

## Why

These files are dead Jito-only patch content. Removing them reduces the maintained fork delta by 718 source lines and 1,496,480 bytes without changing runtime behavior. The current 0.1.10 tip-program binaries remain in place.

## Validation

- repository-wide reference search for the deleted source modules and historical binary filenames
- `cargo check -p solana-program-binaries`


haoranjito (approved):
🔪

PR #1573, opened 2026-08-18
Prune unused Jito dependencies

esemeniuc:
## What changed

- remove unused dependencies from `solana-core`, `bam-banking-bench`, and `bam-local-cluster`
- remove the dead `solana-core` build dependency (the crate has no build script)
- move `bam-banking-bench` to the workspace clap 2 dependency and its equivalent builder APIs
- restore Axum 0.8.8 and remove the Jito-only workspace version pin
- use the existing SHA-2 0.10 line in `solana-bundle`
- regenerate the root, dev-bins, and SBF lockfiles

## Why

The removed manifest entries have no source uses. Their feature/version requirements expanded the Jito dependency graph and lockfiles without providing functionality. The benchmark does not need a special clap 3 requirement; its two options are supported by the workspace clap API.


haoranjito (approved):
:shipit:

PR #1574, opened 2026-08-18
Restore the program cache extract API

esemeniuc:
## What changed

- restore `ProgramCache::extract` to its upstream-compatible arity and cooperative-loading behavior
- add `extract_with_replenish` for the one non-cooperative bundle-simulation path
- return the existing production and test call sites to the default API

## Why

Bundle simulation needs to read cache hits without registering cooperative cache-miss work. Threading that boolean through the common API forced the same literal argument into every other call site and expanded the rebase conflict surface.

## Impact

Behavior is unchanged: the compatibility wrapper always replenishes cooperatively, while `replenish_program_cache_for_simulation` explicitly selects the non-cooperative variant.

## Validation

- `cargo test -p solana-program-runtime loaded_programs::tests --lib` (42 passed)


haoranjito (approved):
:shipit:

PR #1575, opened 2026-08-18
Share Jito validator CLI arguments

esemeniuc:
## What changed

- keep only the eight Jito argument builders shared by startup and runtime admin commands in `commands::jito_args`
- leave startup-only tip, commission, Merkle-authority, and multicast arguments directly at their sole caller
- use one complete help description for each shred argument
- allow both startup and admin shred options to be repeated or supplied as comma-separated entries
- collect every admin occurrence into the existing comma-based RPC payload
- leave BAM in its already-shared feature module, deprecated compatibility handling in `cli.rs`, and generic networking arguments outside the Jito module
- keep `--disable-block-engine-autoconfig` value-less by omitting the incompatible `.value_name(...)`

## Why

The Jito CLI definitions were repeated across validator startup and runtime admin commands. The original cleanup also missed the duplicated retransmit shred receiver argument. Sharing only definitions with multiple callers removes real duplication without introducing one-use wrapper functions.

The admin shred commands previously extracted a single value even when the option appeared more than once. The shared builders now allow repeated one-value occurrences, preserve comma-containing values, and prevent greedy consumption of following arguments.

### DIFF
--- a/validator/src/commands/block_engine/mod.rs
+++ b/validator/src/commands/block_engine/mod.rs
@@ -1,42 +1,19 @@
 use {
-    crate::{admin_rpc_service, cli::DefaultArgs, commands::Result},
-    clap::{App, Arg, ArgMatches, SubCommand, value_t_or_exit},
+    crate::{
+        admin_rpc_service,
+        cli::DefaultArgs,
+        commands::{Result, jito_args},
+    },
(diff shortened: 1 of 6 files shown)


haoranjito (commented):
lgltm. 
please fix `cargo fmt`

haoranjito (approved):

PR #1576, opened 2026-08-18
Minimize Jito SPL fetch overrides

esemeniuc:
## What changed

- restore the upstream `add_spl_program_to_fetch` control flow
- add optional download-base and remote-filename parameters
- express the two Jito program downloads through those narrow overrides

## Why

Jito releases differ in both host/tag layout and remote filename, but that does not require a separate branch through the common SPL registration logic. Keeping the upstream function shape reduces rebase conflicts.

### DIFF
--- a/fetch-spl.sh
+++ b/fetch-spl.sh
@@ -19,29 +19,25 @@ add_spl_program_to_fetch() {
   declare version=$2
   declare address=$3
   declare loader=$4
-  declare param=${5:-}
+  declare artifact=${5:-}
+  declare download_base_url=${6:-}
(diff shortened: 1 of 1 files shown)


haoranjito (approved):
lgtm.
I think the local cluster failure is not relevant. Can we rerun the local clsuter ci to see if it passes now?

esemeniuc:
> lgtm. I think the local cluster failure is not relevant. Can we rerun the local clsuter ci to see if it passes now?

that is upstream flakey Alpenglow test issue unfortunately

PR #1577, opened 2026-08-18
Restore upstream broadcast packet selection

esemeniuc:
## What changed

- restore the upstream iterator pipeline for turbine/next-leader packet selection
- reserve and append external receiver packets after the upstream packet vector is built
- split UDP packets at that append boundary while keeping the complete vector for XDP

## Why

External shred receivers only require a stable split point. Rewriting the upstream selection pipeline as an imperative loop enlarged the conflict surface without changing its policy.

## Impact

Destination order, external-address deduplication/capping, dedicated UDP routing, XDP routing, packet metrics, and first-I/O-error behavior are preserved. Relative to the fork base, this restores 32 upstream lines to their original form.

### DIFF
--- a/turbine/src/broadcast_stage.rs
+++ b/turbine/src/broadcast_stage.rs
@@ -638,52 +638,52 @@ pub fn broadcast_shreds(
         next_broadcast_leader_pubkey(leader_schedule_cache, &working_bank, &my_pubkey, slot)
     };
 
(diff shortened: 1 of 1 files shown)


haoranjito (dismissed):
lgtm. 
it will be good to rerun local cluster.

haoranjito (approved):

PR #1581, opened 2026-08-18
Consolidate bundle RPC test setup

esemeniuc:
## What changed

- add a shared latest-blockhash/rent input helper for bundle RPC cases
- add a default per-transaction bundle simulation config builder
- add one typed RPC response-error extractor
- use struct updates for case-specific config differences

## Why

The bundle integration block repeated the same RPC preamble, six-field config literals, and error destructuring in nearly every case. The helpers keep each case focused on the input or assertion it varies.

## Impact

All case functions and assertions are retained. The integration test is 174 lines smaller.

## Validation

- `cargo test -p solana-rpc-test --test rpc test_simulate_bundle -- --exact`

### DIFF
--- a/rpc-test/tests/rpc.rs
+++ b/rpc-test/tests/rpc.rs
@@ -22,7 +22,9 @@ use {
             RpcBundleExecutionError, RpcBundleSimulationSummary, RpcSimulateBundleConfig,
             SimulationSlotConfig,
         },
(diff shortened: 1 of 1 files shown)


esemeniuc:
Superseded by #1584, which now targets `master` and combines the bundle RPC fixture consolidation with the integration-test relocation so the same test block is reviewed only once.

PR #1582, opened 2026-08-18
Share proxy auth refresh state

esemeniuc:
## What changed

- add `AuthRefreshState` to own the auth client, access token, refresh token, and authenticated validator identity
- centralize identity validation, token refresh, and token replacement
- pass one refresh state through Block Engine and relayer stream setup instead of four separate values
- keep service-specific config checks, counters, datapoint names, select loops, and reconnect behavior local

## Why

The two proxy stages duplicated the security-sensitive refresh core while having genuinely different stream and reconnect policies. Sharing only the auth state removes that duplication without introducing a generic stream driver.

### DIFF
--- a/core/src/proxy/auth.rs
+++ b/core/src/proxy/auth.rs
@@ -27,41 +27,70 @@ pub(crate) struct AuthRefreshState {
 }
 
 impl AuthRefreshState {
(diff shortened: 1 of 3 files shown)


haoranjito (dismissed):

esemeniuc:
fix cargo fmt, sorry!

haoranjito (approved):

PR #1585, opened 2026-08-18
Isolate bundle RPC implementation and test

esemeniuc:
## What changed

- move the bundle simulation request implementation into `rpc/bundles.rs`
- keep the generated JSON-RPC trait declaration and a small delegate in `rpc.rs`
- move the bundle RPC happy-path test alongside the implementation
- build and execute both test requests through one local JSON request/response closure instead of duplicating the request and decoding path
- expose only the test harness fields, methods, helper, and constant required by the sibling test module
- remove imports that became unused in `rpc.rs`

## Why

The bundle-specific implementation and its test occupied roughly 466 lines in the large upstream RPC module. Co-locating them isolates Jito behavior from recurring upstream edits and keeps the test beside the code it covers.

The test previously repeated the complete request literal and response-decoding sequence for the two `replaceRecentBlockhash` cases. The local closure keeps both actual RPC executions and complete response comparisons while removing that duplication.

### DIFF
--- a/rpc/src/rpc.rs
+++ b/rpc/src/rpc.rs
@@ -134,6 +134,7 @@ mod transaction {
 }
 
 pub mod account_resolver;
+mod bundles;
 
 type RpcCustomResult<T> = std::result::Result<T, RpcCustomError>;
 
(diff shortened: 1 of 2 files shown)


haoranjito (dismissed):
lgtm

haoranjito (approved):

PR #1586, opened 2026-08-18
Move bundle simulation bank tests

esemeniuc:
## What changed

- move bundle/program-cache simulation helpers and tests from `bank/tests.rs` into `bank/jito_tests.rs`
- keep test-only setup local to the Jito test module
- remove imports that became unused in the upstream bank test module

## Why

These tests exercise Jito's multi-transaction simulation cache isolation and tombstone behavior. Keeping 365 lines of them in the upstream bank test file increases rebase conflict surface without granting access they cannot get from a sibling child module.

## Impact

All three cases and assertions are retained with fresh setup. This PR is stacked on the bank bundle-simulation module move.

## Validation

- `cargo test -p solana-runtime test_simulate_transactions_unchecked_with_pre_accounts_does_not_populate_global_program_cache --lib`
- `cargo test -p solana-runtime test_simulate_bundle --lib` (2 passed)
- filtered test target rebuilt without warnings after import cleanup


esemeniuc:
Combined into #1583 so the bank bundle-simulation implementation and its Jito-specific tests land as one single-commit PR. The updated #1583 tree exactly matches this PR’s complete head. Closing this superseded stacked PR.

PR #1587, opened 2026-08-18
Move bundle RPC tests with implementation

esemeniuc:
## What changed

- move the in-crate bundle RPC happy-path test alongside `rpc/bundles.rs`
- expose only the test harness type, three used fields, two methods, one helper, and one constant to the sibling child module
- remove test imports that became unused in `rpc.rs`

## Why

The remaining bundle-specific test occupied more than 220 lines in the upstream RPC test module. Co-locating it with the extracted implementation completes the bundle RPC conflict-surface isolation while retaining private harness reuse.

## Impact

The case and assertions are unchanged. This PR is stacked on the bundle RPC implementation move.

## Validation

- `cargo test -p solana-rpc test_rpc_simulate_bundle_happy_path --lib`


esemeniuc:
Combined into #1585 so the bundle RPC implementation and its happy-path test land as one single-commit PR. The updated #1585 tree exactly matches this PR’s complete head. Closing this superseded stacked PR.

PR #1592, opened 2026-08-20
ci: exempt v4.3 from mergify EoL auto-close

0xEdgar:
Mergify's "close PRs against EoL version branches" rule only allowlists v4.0–v4.2, so it auto-closes v4.3 backports like #1591. Mergify evaluates PRs with the config from the PR's base branch, so this fix on `v4.3` is the one that actually stops the closes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -176,6 +176,7 @@ pull_request_rules:
       - base != v4.0
       - base != v4.1
       - base != v4.2
+      - base != v4.3
     actions:
       close:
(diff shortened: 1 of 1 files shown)


0xEdgar:
Mergify closed this PR using the very rule it fixes (config is read from the base branch), so the commit was pushed directly to `v4.3` as 28c2edb04c. Master sync is #1593.

PR #1596, opened 2026-08-21
banking-stage: restore votes after bank replacement

esemeniuc:
## Summary

- retain taken vote bytes with the leader bank identity
- restore eligible votes when a different bank replaces the leader bank at the same slot
- preserve external duplicate suppression and discard stale, incompatible, or superseded retained votes
- restore votes before the VoteWorker empty-storage early return

## Issue
During a sad Alpenglow handover, the provisional leader bank can be cleared after VoteWorker has removed a recorded vote from VoteStorage. BlockCreationLoop reinjects recorded transactions through the ordinary scheduler, which does not schedule while BAM is connected. Retaining the taken vote inside VoteStorage allows it to be restored directly when the same-slot replacement bank appears.


haoranjito:
I am not familiar with the new vote process for alpenglow. so I am not sure that what need to happen here. 
Why this change need to be in jito-solana? Should it be in agave upstream?

0xEdgar (approved):

PR #1602, opened 2026-08-24
ci: Build and publish release artifacts on tag push

0xEdgar:
Linear: MEV-12300
Buildkite agave-secondary stopped publishing tarballs around v3.0
(release.jito.wtf is stale since Dec 2024), so deploys rebuild from
source every time. Run ./f on tag pushes and manual dispatch, then
attach tarball, version manifest, and SHA-256 to the GitHub release
so Ansible can download-and-swap by exact version (MEV-12300).

- Dispatch refs naming an existing release backfill its assets
- Other refs are kept as workflow artifacts
- Stock runner disk is too small; drop unused toolchains and move
  the docker data-root to /mnt

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/release-artifacts.yml
+++ b/.github/workflows/release-artifacts.yml
@@ -126,3 +126,4 @@ jobs:
         with:
           name: jito-solana-release
           path: dist/
+          compression-level: 0  # contents are already bzip2
--- a/.github/workflows/release-artifacts.yml
+++ b/.github/workflows/release-artifacts.yml
@@ -11,6 +11,8 @@ on:
   pull_request:
     paths:
       - ".github/workflows/release-artifacts.yml"
+      - "f"
+      - "dev/Dockerfile"
   workflow_dispatch:
     inputs:
       ref:
@@ -86,12 +88,16 @@ jobs:
 
       - name: Checksum and summarize
         run: |
+          ./docker-output/agave-validator --version
           cd dist
           for f in *.tar.bz2; do
             sha256sum "$f" > "$f.sha256"
           done
           {
             echo "## jito-solana release artifact"
+            echo '```'
+            ../docker-output/agave-validator --version
+            echo '```'
             echo '```'
             cat ./*.yml
             echo '```'
--- a/.github/workflows/release-artifacts.yml
+++ b/.github/workflows/release-artifacts.yml
@@ -18,6 +18,9 @@ on:
         required: true
         description: git ref to build (tag, branch, or SHA)
 
+permissions:
+  contents: read
+
 concurrency:
   group: release-artifacts-${{ github.event.pull_request.number || inputs.ref || github.ref_name }}
   cancel-in-progress: ${{ github.event_name == 'pull_request' }}
@@ -73,7 +76,11 @@ jobs:
           fi
           if [[ "$EVENT" == "workflow_dispatch" ]]; then
             git fetch --depth=1 origin "$DISPATCH_REF"
-            args+=(--checkout "$(git rev-parse FETCH_HEAD)")
+            args+=(--checkout "$(git rev-parse 'FETCH_HEAD^{commit}')")
+            if [[ -z "$TAG" ]]; then
+              # keep the human-readable ref in the artifact name
+              args+=(--tag "${DISPATCH_REF//\//_}-$(git rev-parse --short 'FETCH_HEAD^{commit}')")
+            fi
           fi
           ./f "${args[@]}"
 
--- a/.github/workflows/release-artifacts.yml
+++ b/.github/workflows/release-artifacts.yml
@@ -1,18 +1,13 @@
-# Build the versioned jito-solana validator artifact in CI and publish it so
-# deploys can download-and-swap instead of building from source (MEV-12300).
-#
-# Drives ./f (dev/Dockerfile), which produces a hermetic linux/amd64 release
-# tarball + version manifest. Tag pushes (and dispatch refs that name an
-# existing GitHub release) upload the tarball, manifest, and SHA-256 to that
-# release; any other dispatch ref is kept as a workflow artifact.
+# Build the release tarball with ./f and attach it (+ manifest, SHA-256) to
+# the tag's GitHub release, so deploys download instead of building from
+# source (MEV-12300).
 name: Release Artifacts
 
 on:
   push:
     tags:
       - "v*-jito"
-  # Self-test: PRs touching this workflow run the full build pre-merge.
-  # PR runs never touch a release (artifact-upload path only).
+  # Self-test: PR runs build but never touch a release.
   pull_request:
     paths:
       - ".github/workflows/release-artifacts.yml"
@@ -29,12 +24,11 @@ concurrency:
 
 jobs:
   build:
-    # Fork PRs get a read-only token (GHCR login would fail); skip them.
+    # Skip fork PRs: their token cannot push the GHCR cache.
     if: >-
       github.repository == 'jito-foundation/jito-solana' &&
       (github.event_name != 'pull_request' ||
        github.event.pull_request.head.repo.full_name == github.repository)
-    # Org larger runner: 16-core / 64GB / 600GB SSD (label big-runner-1).
     runs-on: big-runner-1
     timeout-minutes: 360
     permissions:
@@ -47,10 +41,8 @@ jobs:
           ref: ${{ inputs.ref || github.ref }}
           submodules: recursive
 
-      # Tag pushes always name the artifact after the tag. A dispatch ref gets
-      # the release name only when a GitHub release for that exact ref already
-      # exists (backfills); anything else keeps ./f's derived
-      # v<version>_<shortsha> naming.
+      # Empty tag = not a release build; artifact keeps ./f's derived name
+      # and is uploaded as a workflow artifact instead.
       - name: Resolve release tag
         id: meta
         env:
@@ -68,11 +60,8 @@ jobs:
           fi
           echo "tag=$tag" >> "$GITHUB_OUTPUT"
 
-      # Registry layer cache skips the toolchain stage (apt + rustup) on
-      # warm runs. It does NOT cache the cargo compile: that state lives in
-      # BuildKit cache mounts, which registry cache export doesn't cover.
-      # Registry cache export also requires a docker-container builder (the
-      # default docker driver can't export cache).
+      # Layer cache only (toolchain stage); cargo state lives in BuildKit
+      # cache mounts, which registry cache cannot export.
       - name: Set up registry build cache
         env:
           GH_TOKEN: ${{ github.token }}
@@ -91,8 +80,6 @@ jobs:
           fi
           ./f "${args[@]}"
 
-      # Publish the source revision, checksum, and contents alongside the
-      # artifact (see the job summary and the .sha256 asset).
       - name: Checksum and summarize
         run: |
           cd dist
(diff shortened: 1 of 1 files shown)


.github/workflows/release-artifacts.yml
```diff
@@ -0,0 +1,129 @@
...
+            echo '{"data-root": "/mnt/docker"}' \
+              | sudo tee /etc/docker/daemon.json >/dev/null
+          fi
+          sudo systemctl start docker
+          df -h / /mnt
+
+      - name: Checkout
+        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0  # v7.0.0
+        with:
+          ref: ${{ inputs.ref || github.ref }}
```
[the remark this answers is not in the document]
0xEdgar: Fixed in a9fb16e635 — dispatch builds now keep the workflow's checkout and pass the requested ref to `./f --checkout`, so old tags (whose `f` predates the cache flags, i.e. ≤v4.1.x) build with current tooling.

.github/workflows/release-artifacts.yml
```diff
@@ -0,0 +1,137 @@
...
+    timeout-minutes: 360
+    permissions:
+      contents: write
+      packages: write
+    steps:
+      - name: Checkout
+        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0  # v7.0.0
+        with:
+          ref: ${{ inputs.ref || github.ref }}
+          submodules: recursive
```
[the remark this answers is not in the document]
0xEdgar: Fixed in a9fb16e635 — verified the premise: `f` before v4.2.0-jito lacks `--pull-cache/--push-cache`. Dispatch now fetches the ref and builds it via `./f --checkout` from the workflow's own checkout.

.github/workflows/release-artifacts.yml
```diff
@@ -0,0 +1,129 @@
...
+              | sudo tee /etc/docker/daemon.json >/dev/null
+          fi
+          sudo systemctl start docker
+          df -h / /mnt
+
+      - name: Checkout
+        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0  # v7.0.0
+        with:
+          ref: ${{ inputs.ref || github.ref }}
+          submodules: recursive
```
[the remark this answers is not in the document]
0xEdgar: Fixed in a9fb16e635 — `persist-credentials: false` added; no step needs the persisted credential (later `gh`/fetch steps use explicit `GH_TOKEN` or anonymous HTTPS on this public repo).

.github/workflows/release-artifacts.yml
```diff
@@ -0,0 +1,137 @@
...
+    if: >-
+      github.repository == 'jito-foundation/jito-solana' &&
+      (github.event_name != 'pull_request' ||
+       github.event.pull_request.head.repo.full_name == github.repository)
+    # Org larger runner: 16-core / 64GB / 600GB SSD (label big-runner-1).
+    runs-on: big-runner-1
+    timeout-minutes: 360
+    permissions:
+      contents: write
+      packages: write
```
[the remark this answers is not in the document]
0xEdgar: Declining the job split: the pull_request trigger executes the PR's own modified workflow file, so a same-repo PR controls its permissions block regardless — a read-only build job wouldn't create a real boundary. Same-repo PR authors are committers who can already push tags/branches directly; fork PRs (the actual untrusted path) are skipped. Credential persistence removed in a9fb16e635 as extra hardening.

.github/workflows/release-artifacts.yml
```diff
@@ -0,0 +1,115 @@
...
+          DISPATCH_REF: ${{ inputs.ref }}
+          TAG: ${{ steps.meta.outputs.tag }}
+        run: |
+          args=()
+          if [[ -n "$TAG" ]]; then
+            args+=(--tag "$TAG")
+          fi
+          if [[ "$EVENT" == "workflow_dispatch" ]]; then
+            git fetch --depth=1 origin "$DISPATCH_REF"
+            args+=(--checkout "$(git rev-parse FETCH_HEAD)")
```
jito-spaceman: May need to be careful here on specific trigger paths, i.e., on a ref with no existing release this may tag the build as `<fullsha>-<shortsha>` instead of a readable ref. I think we should maybe pass the human-readable ref through as well, for example by adding `--tag "${DISPATCH_REF//\//_}-$(git rev-parse --short FETCH_HEAD)"` when TAG is empty. Probably needs tested to be certain.

.github/workflows/release-artifacts.yml
```diff
@@ -0,0 +1,115 @@
...
+  pull_request:
+    paths:
+      - ".github/workflows/release-artifacts.yml"
+  workflow_dispatch:
+    inputs:
+      ref:
+        type: string
+        required: true
+        description: git ref to build (tag, branch, or SHA)
+
```
jito-spaceman, line 22: Should probably add here

```
permissions:
  contents: read
```

.github/workflows/release-artifacts.yml
```diff
@@ -0,0 +1,115 @@
...
+            echo '```'
+            cat ./*.sha256
+            echo '```'
+            echo "### Contents"
+            echo '```'
+            tar -tjf ./*.tar.bz2
+            echo '```'
+          } >> "$GITHUB_STEP_SUMMARY"
+          ls -lh
+
```
jito-spaceman, line 124: Should we add a smoketest here to dump the `agave-validator --version`?

jito-spaceman: May need to run in the docker though to leverage required deps

.github/workflows/release-artifacts.yml
0xEdgar: Done in 854154388c — when a dispatch ref has no release, the build now passes `--tag "${DISPATCH_REF//\//_}-$(git rev-parse --short 'FETCH_HEAD^{commit}')"` so the artifact keeps the readable ref name (peeled to the commit so annotated tags short-sha consistently with ./f's own naming).

.github/workflows/release-artifacts.yml
0xEdgar: Done in 854154388c — workflow-level `permissions: contents: read`; the build job still elevates to write for the release upload.

.github/workflows/release-artifacts.yml
0xEdgar: Done in 729608ec88 — `./docker-output/agave-validator --version` runs after the build (loose binaries are already exported, no extraction needed) and the output lands in the job summary.

.github/workflows/release-artifacts.yml
```diff
@@ -0,0 +1,123 @@
...
+name: Release Artifacts
+
+on:
+  push:
+    tags:
+      - "v*-jito"
+  # Self-test: PR runs build but never touch a release.
+  pull_request:
+    paths:
+      - ".github/workflows/release-artifacts.yml"
```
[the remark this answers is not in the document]
0xEdgar: Done in 729608ec88 — `f` and `dev/Dockerfile` added to the self-test paths filter.

.github/workflows/release-artifacts.yml
```diff
@@ -0,0 +1,135 @@
...
+      - name: Publish to GitHub release
+        if: steps.meta.outputs.tag != ''
+        env:
+          GH_TOKEN: ${{ github.token }}
+          TAG: ${{ steps.meta.outputs.tag }}
+        run: |
+          # release.yml normally creates the release on tag push
+          if ! gh release view "$TAG" --repo "$GITHUB_REPOSITORY" >/dev/null 2>&1; then
+            gh release create "$TAG" --repo "$GITHUB_REPOSITORY" \
+              --prerelease --title "Release $TAG" --notes "🚧"
```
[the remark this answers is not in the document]
0xEdgar: Resolved by deletion in f0b061914f — the create fallback is gone; release.yml solely owns creation and the upload fails loudly if the release is absent.

.github/workflows/release-artifacts.yml
```diff
@@ -0,0 +1,128 @@
...
+        if: steps.meta.outputs.tag != ''
+        env:
+          GH_TOKEN: ${{ github.token }}
+          TAG: ${{ steps.meta.outputs.tag }}
+        # release.yml creates the release on tag push; fail if it is missing
+        run: gh release upload "$TAG" dist/* --clobber --repo "$GITHUB_REPOSITORY"
+
+      - name: Upload workflow artifact
+        if: steps.meta.outputs.tag == ''
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a  # v7.0.1
```
jito-spaceman, line 135: `upload-artifact` zips by default, add `compression-level: 0` so we don't zip a bzip2 unnecessarily

0xEdgar: Done in e66749addb.

.github/workflows/release-artifacts.yml
```diff
@@ -0,0 +1,129 @@
...
+          fi
+          if [[ "$EVENT" == "workflow_dispatch" ]]; then
+            git fetch --depth=1 origin "$DISPATCH_REF"
+            args+=(--checkout "$(git rev-parse 'FETCH_HEAD^{commit}')")
+            if [[ -z "$TAG" ]]; then
+              # keep the human-readable ref in the artifact name
+              args+=(--tag "${DISPATCH_REF//\//_}-$(git rev-parse --short 'FETCH_HEAD^{commit}')")
+            fi
+          fi
+          ./f "${args[@]}"
```
jito-spaceman, line 96: Might need another gate (untested) to properly avoid the edge case that would dump a `<40-char>-<7-char>` SHA pattern:

```suggestion
          if [[ "$EVENT" == "workflow_dispatch" ]]; then
            git fetch --depth=1 origin "$DISPATCH_REF"
            sha="$(git rev-parse 'FETCH_HEAD^{commit}')"
            short="$(git rev-parse --short "$sha")"
            args+=(--checkout "$sha")
            if [[ -z "$TAG" ]]; then
              # Keep the human-readable ref in the artifact name. A SHA ref
              # is already the commit, so use it alone.
              name="${DISPATCH_REF//\//_}"
              if [[ "$sha" == "$name"* ]]; then
                name="$short"
              else
                name="${name}-${short}"
              fi
              args+=(--tag "$name")
            fi
          fi
```

.github/workflows/release-artifacts.yml
```diff
@@ -0,0 +1,129 @@
...
+            sha256sum "$f" > "$f.sha256"
+          done
+          {
+            echo "## jito-solana release artifact"
+            echo '```'
+            ../docker-output/agave-validator --version
+            echo '```'
+            echo '```'
+            cat ./*.yml
+            echo '```'
```
jito-spaceman, line 113: Back-to-back code blocks reads kind of messy. Can we clean up or just add an additional header to clean it up a bit?

.github/workflows/release-artifacts.yml
0xEdgar: Done in 6f6fa402a3 — applied the suggested gate: a dispatch ref that is a SHA prefix of the resolved commit names the artifact by short SHA alone.

.github/workflows/release-artifacts.yml
0xEdgar: Done in 6f6fa402a3 — added a "Version manifest" header between the version and manifest code blocks.

.github/workflows/release-artifacts.yml
```diff
@@ -0,0 +1,139 @@
...
+          GH_TOKEN: ${{ github.token }}
+          EVENT: ${{ github.event_name }}
+          PUSH_TAG: ${{ github.ref_name }}
+          DISPATCH_REF: ${{ inputs.ref }}
+        run: |
+          tag=""
+          if [[ "$EVENT" == "push" ]]; then
+            tag="$PUSH_TAG"
+          elif [[ -n "$DISPATCH_REF" ]] \
+            && gh release view "$DISPATCH_REF" --repo "$GITHUB_REPOSITORY" >/dev/null 2>&1; then
```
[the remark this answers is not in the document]
0xEdgar: will address in a follow up.

jito-spaceman (approved):
let's go

PR #1607, opened 2026-08-28
TipRouter Snapshot Service

hash-envy:
## Problem

 The Tip Router needs information about tip-distribution accounts and staking metadata, from the exact epoch-boundary `Bank` each epoch. 
 
## Summary of Changes

  - Adds a jito-tip-router-snapshot-service workspace crate that captures and generates stake metadata "candidates" at epoch boundaries. "Candidates" are frozen Banks that have not yet been rooted. This is a precaution in case of forking at the boundary
  - Utilizes the `BankNotificationBroadcaster` introduced in #1554 to read `FrozenBank` + `RootedChain` notifications. These notifications drive the state-machine
  - If a `FrozenBank` is an epoch-boundary candidate, an artifact worker thread is launched, and the StakeMeta artifact is created
  - When the boundary slot becomes rooted, the JSON artifact is atomically published to a specific directory 
  - Integrates the service into validator startup behind new --enable-tip-router-snapshot-service configuration and requires --no-voting.

## Notes
- Pretty much all of this code is opt-in based on --enable-tip-router-snapshot-service and will only be run by Tip Router Operators. It should not affect perf of other parts of the system otherwise
- The `stake-meta` calculation + capture is an optimized version of what previously existed in tip-router. It is optimized for reduced "Bank-Hold" times (~15ms). In testing it has produced identical artifacts as the legacy version across many epochs
- Most of the flow logic exists in `jito-tip-router-snapshot-service/src/service/context.rs`. The "context" is the main home of the core state machine. `service/mod.rs` drives the context / state-machine.
- `service/publication_state.rs` contains info on multiple forks and `service/worker_pool` manages the artifact-creation threads. `snapshot_worker.rs` handles the actual logic / work of snapshotting.


jito-tip-router-snapshot-service/Cargo.toml
```diff
@@ -0,0 +1,36 @@
...
+[dependencies]
+
+bincode = { workspace = true }
+borsh = { workspace = true }
+clap = { workspace = true }
+crossbeam-channel = { workspace = true }
+jito-priority-fee-distribution-sdk = { git = "https://github.com/jito-foundation/jito-tip-router.git", rev = "d3e10e9ca650197a78b01f64c492b1cd91eea7f4" }
+jito-stake-meta-types = { git = "https://github.com/jito-foundation/jito-tip-router.git", rev = "d3e10e9ca650197a78b01f64c492b1cd91eea7f4" }
+jito-tip-distribution-sdk = { git = "https://github.com/jito-foundation/jito-tip-router.git", rev = "d3e10e9ca650197a78b01f64c492b1cd91eea7f4" }
+jito-tip-payment-sdk = { git = "https://github.com/jito-foundation/jito-tip-router.git", rev = "d3e10e9ca650197a78b01f64c492b1cd91eea7f4" }
```
ebatsell, line 16: Calling out for others - these sdks from tip router are quite static and minimal if for some reason we wanted to remove these dependencies and copy them here

ebatsell (commented):
Tip-router specific things look good, we know it's been creating the same merkle roots for a while now. And from my eyes the shape of the service is clean and follows patterns nicely but will defer to jito-solana owners to give input and final approval

Cargo.toml
```diff
@@ -272,6 +273,7 @@ jemallocator = { package = "tikv-jemallocator", version = "0.6.0", features = [
     "stats",
 ] }
 jito-protos = { path = "jito-protos", version = "=4.4.0-alpha.2" }
+jito-tip-router-snapshot-service = { path = "jito-tip-router-snapshot-service", version = "=4.4.0-alpha.2" }
```
0xEdgar, line 276: wdyt about putting this crate behind a `--features tip-router` flag? reasoning: 99.9% of validators aren't going to need to compile this code.

core/src/validator.rs
```diff
@@ -740,6 +749,7 @@ pub struct Validator {
     /// in sync with kernel route availability.
     root_multicast_shred_check_service: Option<MulticastShredCheckService>,
     sample_performance_service: Option<SamplePerformanceService>,
+    tip_router_snapshot_service: Option<TipRouterSnapshotService>,
```
0xEdgar: i think it might be better for core to not really need to know about tip-router, and instead accepts an extra subscriber as an owned argument. keeps core agnostic to tip-router

jito-tip-router-snapshot-service/src/notification_filter.rs
```diff
@@ -0,0 +1,23 @@
+use solana_rpc::optimistically_confirmed_bank_tracker::{BankNotification, BankNotificationFilter};
+
+/// Accepts epoch-boundary frozen banks and exact rooted chains the Tip Router service can act on.
+#[derive(Clone, Copy, Debug, Default)]
+pub struct TipRouterEpochBoundaryFilter;
+
+impl BankNotificationFilter for TipRouterEpochBoundaryFilter {
```
0xEdgar, line 19: wonder if we're able to add some unit tests for this

0xEdgar: clanker suggestion:
--------
Suggest pulling the check into a pure fn, e.g. `is_epoch_boundary_child(epoch, parent_slot, &EpochSchedule) -> bool`, so most cases don't need a Bank. Then:

An exhaustive sweep over all (parent, child) slot pairs on a small schedule (EpochSchedule::custom(32, 32, false) and once with warmup = true), asserting against get_epoch(child) > get_epoch(parent).
A couple of Bank tests for the wiring: Bank::new_for_tests + Bank::new_from_parent(parent, SlotLeader::default(), slot). The channel_with_filter tests in rpc/src/optimistically_confirmed_bank_tracker.rs (~L696-860) are a good template.

Cases worth naming explicitly: skipped leading slots of N+1, skipped trailing slots of N (candidate is not the last slot of the epoch), skipped entire epoch (parent in N, child in N+2, so the candidate epoch is N), genesis bank (parent_slot == 0), and the non-Frozen variants.

One extra assertion that would be nice: for each Frozen case, eligible_candidate_from_boundary_child returns Some iff the filter returns true, since the filter uses parent_slot() and the context uses parent().

jito-tip-router-snapshot-service/src/service/mod.rs
```diff
@@ -0,0 +1,149 @@
...
+            };
+            TipRouterSnapshotServiceError::ArtifactStoreInitialization { path, source }
+        })?;
+
+        let thread_hdl = Builder::new()
+            .name("tipRtSnapshot".to_string())
+            .spawn(move || {
+                info!("TipRouterSnapshotService has started");
+                let result = Self::run(config, artifact_store, bank_notification_receiver, exit);
+                if let Err(e) = result.as_ref() {
```
0xEdgar, line 87: wonder if we should distinguish fatal from recoverable errors here. otherwise we could have a non-fatal epoch error until someone notices the file isn't there. also prob should emit datapoint here for observability

jito-tip-router-snapshot-service/src/service/context.rs
```diff
@@ -0,0 +1,258 @@
...
+        match notification {
+            BankNotification::Frozen(bank) => self.handle_frozen_bank(config, artifact_store, bank),
+            BankNotification::NewRootedChain(rooted_chain, _) => {
+                self.handle_new_rooted_chain(rooted_chain, artifact_store)
+            }
+            _ => Ok(()),
+        }
+    }
+
+    fn handle_new_rooted_chain(
```
0xEdgar, line 62: we're not checking that the worker reports Written, meaning we could race the File not existing yet (causing us to miss the epoch), or file is being written to. fix would be to wait for both rooted + file written. also prob worth emitting here for completeness sake.

0xEdgar (commented):
thanks for the change, some comments. lmk what you think. will do another pass shortly.

Cargo.toml
hash-envy: I considered this, but you end up with a ton of `#[cfg(feature...)]` littered all around the validator code --everywhere there is a branch on the existence of the cli flag. It's sort of redundant. 

More importantly, we would have to publish separate releases each time. That might be a pain

I'm still open to it though. These are just worth considering

jito-tip-router-snapshot-service/src/service/context.rs
hash-envy: The idea was that rooting is always 64 slots after freezing, so this should never happen. Alpenglow might be changing this though...

jito-tip-router-snapshot-service/src/service/mod.rs
hash-envy: Yeah i guess a lot of these errors, we can give up on the current epoch and start trying for the next. This would be better than just shutting off the service. 

Good call on the datapoint. That's probably the best way for observability and alerting rn.

Cargo.toml
0xEdgar: yea the main concern is the cfg count. if core takes extra subscribers (my other comment), core would have 0 cfgs and feature exists only in agave validator

fable implement it make no mistakes
```rust
// validator/src/commands/run/tip_router.rs
#[cfg(feature = "tip-router")]
mod imp {
    pub fn args() -> Vec<Arg> { cli::args() }
    pub fn setup(matches, senders: &mut Vec<BankNotificationSender>) -> Result<Option<Pending>> { .. }
    pub fn start(pending, exit) -> Result<Option<Service>> { .. }
    pub fn join(service) { .. }
}
#[cfg(not(feature = "tip-router"))]
mod imp {
    pub fn args() -> Vec<Arg> { vec![] }
    pub fn setup(..) -> Result<Option<Pending>> { Ok(None) }
    pub fn start(..) -> Result<Option<Service>> { Ok(None) }
    pub fn join(..) {}
}
pub use imp::*;
```

for the releases: i think thats a fairly small change. we already handle tiprouter deployments separately, would be a small addition to `./f` to accept a feature gate.

core/src/validator.rs
0xEdgar:
```
/ core/src/validator.rs — no tip-router knowledge, no cfg
pub fn new_with_exit(..., extra_bank_notification_senders: Vec<BankNotificationSender>) -> ... {
    ...
    let mut senders = extra_bank_notification_senders;
    if rpc_enabled { senders.insert(0, rpc_sender); }
    BankNotificationSenderConfig {
        sender: BankNotificationBroadcaster::new(senders),
        should_send_parents: geyser || has_extra,
        ...
    }
    ...
}

// validator/src/commands/run/tip_router.rs — the only cfgs in the tree
#[cfg(feature = "tip-router")]
mod imp {
    pub fn args() -> Vec<Arg> { ... }
    pub fn setup(matches, senders: &mut Vec<BankNotificationSender>) -> Result<Option<Pending>> { ... }
    pub fn start(pending, exit) -> Result<Option<Service>> { ... }
    pub fn join(service) { ... }
}
#[cfg(not(feature = "tip-router"))]
mod imp { /* same signatures, all no-ops */ }
pub use imp::*;
```

jito-tip-router-snapshot-service/src/notification_filter.rs
hash-envy: Done

jito-tip-router-snapshot-service/src/service/context.rs
hash-envy: Done. Now the publish trigger is one of:
- WriteWorkerFinishedNotification (if already rooted)
- RootedNotification (if candidate worker has completed)

jito-tip-router-snapshot-service/src/service/mod.rs
hash-envy: I ended up just adding metrics for tracking failures. In most failure cases, the candidate write is completed, but not published. We can manually publish this if there is indeed ever an error

Cargo.toml
hash-envy: Ah. This is a good suggestion. Thank you. I've implemented it

core/src/validator.rs
hash-envy: Yeah. Per my later reply to the config-based comment, i've added this. It's a good point

jito-tip-router-snapshot-service/src/notification_filter.rs
hash-envy: Done

.github/workflows/release-artifacts.yml
```diff
@@ -29,11 +29,22 @@ concurrency:
...
 jobs:
   build:
+    name: Build ${{ matrix.variant }} artifact
     # Skip fork PRs: they cannot use org larger runners and would queue forever.
     if: >-
       github.repository == 'jito-foundation/jito-solana' &&
       (github.event_name != 'pull_request' ||
        github.event.pull_request.head.repo.full_name == github.repository)
+    strategy:
+      fail-fast: false
```
0xEdgar, line 39: hmm maybe set to true?

hash-envy: The thinking was that i didnt want tip-router blocking a release especially in a time-crunch

0xEdgar (approved):

PR #1617, opened 2026-09-03
MEV-12682: gate BAM work until leader bank resolution

esemeniuc:
MEV-12682 / JSA-24: during fast Alpenglow handover, BAM could validate or execute against a provisional bank and return a terminal result for work that should run on its replacement.

Hold both atomic and ordinary BAM batches until the leader bank resolves. Capture the parser bank and resolution flag together, deferring blockhash/status-cache and seed fee-payer checks while unresolved. Use one resolved, active bank matching the scheduler slot for both scheduler validation passes and dispatch; this also removes both scheduler BankForks reads.

Reuse master's retained bank_slot and Hold handling through replacement gaps. Remove the obsolete atomic-only deferred queue. Regression coverage includes deferred parser failures, selected-bank validation, stale Consume during replacement, and both sad-handover variants with provisional rejection and a successful control transaction.

Validation: 55 BAM, scheduler-controller, and block-creation-loop tests passed. Pinned-nightly Clippy (--lib --tests -- -D warnings) and workspace formatting passed.

### DIFF
--- a/core/src/banking_stage.rs
+++ b/core/src/banking_stage.rs
@@ -719,7 +719,6 @@ impl BankingStage {
                             work_sender,
                             finished_work_receiver,
                             bam_dependencies.outbound_sender.clone(),
-                            bam_scheduler_bank_forks.clone(),
                             bam_shared_leader_state.clone(),
                         );
                         let receive_and_buffer = BamReceiveAndBuffer::new(
--- a/core/src/banking_stage/transaction_scheduler/bam_receive_and_buffer.rs
+++ b/core/src/banking_stage/transaction_scheduler/bam_receive_and_buffer.rs
@@ -210,7 +210,7 @@ impl BamReceiveAndBuffer {
                 }
             }
 
-            let working_bank = loop {
+            let (working_bank, bank_checks_enabled) = loop {
                 if exit.load(Ordering::Relaxed) || shutdown.load(Ordering::Relaxed) {
                     return;
                 }
@@ -281,6 +281,7 @@ impl BamReceiveAndBuffer {
                                 max_schedule_slot,
                                 (&root_bank, &working_bank),
                                 &blacklisted_accounts,
+                                bank_checks_enabled,
                                 &mut metrics,
                             ));
                         stats.accumulate(parse_stats);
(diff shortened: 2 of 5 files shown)


haoranjito:
I am not sure if I understand the problem. 

`another bank with the same numeric slot` I though that slot will monotonically increasing with leader scheduler. each slot can only be taken by one leader? 

Also, can you explain what's the atomic_batches_enabled signal and new identifying_pending_bank signal do?

core/src/banking_stage/decision_maker.rs
```diff
@@ -65,14 +66,20 @@ impl DecisionMaker {
             } else {
                 BufferedPacketsDecision::Consume(working_bank.clone())
             }
+        } else if identify_pending_bank && !state.atomic_batches_enabled() {
```
haoranjito: ok. I understand the change now. 

for bam, when there is no working bank and atomic_batches_enabled is false, this means we are in the bank gap of the sad path of alpenglow. so we should do forward and hold.

core/src/banking_stage/decision_maker.rs
```diff
@@ -65,14 +66,20 @@ impl DecisionMaker {
...
+            BufferedPacketsDecision::ForwardAndHold
         } else if let Some(leader_first_tick_height) = state.leader_first_tick_height() {
             let current_tick_height = state.tick_height();
             let ticks_until_leader = leader_first_tick_height.saturating_sub(current_tick_height);
             if ticks_until_leader
                 <= (FORWARD_TRANSACTIONS_TO_LEADER_AT_SLOT_OFFSET - 1) * DEFAULT_TICKS_PER_SLOT
             {
                 BufferedPacketsDecision::Hold
-            } else if ticks_until_leader < HOLD_TRANSACTIONS_SLOT_OFFSET * DEFAULT_TICKS_PER_SLOT {
+            } else if !identify_pending_bank
```
haoranjito: we do leader soon buffer and hold only for non-bam. don't need that for bam. 

This is an optimization, we could do that for bam too. not directly relevant for current fix. right?

esemeniuc: larger leader-soon buffering for bam could be a separate optimizatio, but this `if` is needed. before, bam returned batches for both Forward and ForwardAndHold. 

this pr makes ForwardAndHold keep batches during the replacement gap, so we use Forward for BAM in the ordinary leader-soon case, preserving its previous behavior.

0xEdgar:
any idea on what the lag this introduces will be? from my understanding here previously BAM txns would start executing before bank resolves (parent notarized). somewhat worried it could lead to us packing smaller blocks

0xEdgar (approved):

PR #1618, opened 2026-09-03
banking_stage: preserve BAM cost admission order (JSA-72)

esemeniuc:
MEV-12649 / JSA-72: independent BAM batches share eight workers, so a later batch could reserve the remaining block budget before an earlier batch. Near the block limit, this rejected earlier work even when both batches would fit after cost settlement.

Reserve costs on the selected resolved Bank in BAM priority-pop order. Defer a block-limit failure when earlier unsettled estimates could cover the shortfall, then retry the deferred head before later work. Workers consume admission on its exact Bank; returned or unsent work releases its reservation exactly once. Allocate admission results before locking the cost tracker.

On same-slot Bank replacement, workers return stale reservations to the scheduler. The scheduler releases them on the owning Bank, drains outstanding work, and retries returned batches in original dispatch order using stable batch IDs.

Prepare required tip programs before the first BAM cost admission on each concrete Bank, retaining queued work on preparation failure. Reconcile replacement Banks before the idle shortcut; deferred admissions remain eligible for retry even when the priority graph is empty. Preparation is cached per Bank; refreshing builder metadata after preparation on the same Bank remains outside this change.

Tip maintenance runs only on the scheduler. Remove the unused worker fallback, its alternate constructor and dependency plumbing, and the tip-only mutex. Tip tests own their preparation dependencies directly; account locking and reservation handling are preserved.

Includes the bank-resolution work merged in #1617, including deferred fork checks when a provisional Bank has not yet been published in shared leader state.

Validation after worker tip cleanup: 263 banking-stage and block-creation-loop tests passed, including unpublished/resolved-Bank validation, JSA-72 priority inversion, deferred-head retries, reservation refunds, tip preparation/recovery, and repeated bank replacement. Pinned-nightly Clippy (`--lib --tests -- -D warnings`) and workspace formatting passed. The quiet-host performance comparison remains open.

### DIFF
--- a/core/src/banking_stage/transaction_scheduler/bam_scheduler.rs
+++ b/core/src/banking_stage/transaction_scheduler/bam_scheduler.rs
@@ -370,6 +370,7 @@ impl<Tx: TransactionWithMeta> BamScheduler<Tx> {
                         )
                     }),
                 self.inflight_reserved_cost,
+                revert_on_error,
             );
             let Some((results, reserved_cost)) = attempt else {
                 debug!(
@@ -1950,7 +1951,7 @@ mod tests {
     }
 
     #[test_case::test_case(1; "single_transaction")]
-    #[test_case::test_case(2; "partial_batch_rollback")]
+    #[test_case::test_case(2; "atomic_prefix_rollback")]
     fn test_deferred_batch_is_final_once_inflight_settles_without_freeing_budget(
         second_batch_size: usize,
     ) {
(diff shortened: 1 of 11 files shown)


core/src/banking_stage/transaction_scheduler/bam_scheduler.rs
```diff
@@ -205,13 +229,62 @@ impl<Tx: TransactionWithMeta> BamScheduler<Tx> {
...
+                }
+                self.pending_admission.pop_first();
+                (TransactionBatchId::new(batch_id), id)
+            } else {
+                let Some(id) = self.prio_graph.pop() else {
+                    return Ok(num_scheduled);
+                };
+                (self.get_next_schedule_id(), id)
+            };
+
```
0xEdgar, line 287: could this lead to unused capacity near slot the deadline? big deferred batch blocks smaller independent batches that previously would immediately fit

esemeniuc: updated

0xEdgar (approved):

PR #1623, opened 2026-09-04
ci: publish release artifacts for -jito.N hotfix tags

0xEdgar:
problem:

The release-artifacts workflow triggers on the tag glob `v*-jito`, which does not match `-jito.1` , which we want to cut during patch releases.

fix: match `*jito.[0-9]*`

### DIFF
--- a/.github/workflows/release-artifacts.yml
+++ b/.github/workflows/release-artifacts.yml
@@ -7,6 +7,7 @@ on:
   push:
     tags:
       - "v*-jito"
+      - "v*-jito.[0-9]*"  # hotfix re-cuts, e.g. v4.2.2-jito.1
   # Self-test: PR runs build but never touch a release.
   pull_request:
     paths:


jito-spaceman (approved):

.github/workflows/release-artifacts.yml
```diff
@@ -7,6 +7,7 @@ on:
   push:
     tags:
       - "v*-jito"
+      - "v*-jito.[0-9]*"  # hotfix re-cuts, e.g. v4.2.2-jito.1
```
esemeniuc, line 10: With `*`, a tag like `v4.2.2-jito.1abc` also matches.

```suggestion
      - "v*-jito.[0-9]+"  # hotfix re-cuts, e.g. v4.2.2-jito.1
```

jito-spaceman: I approved on the premise that we already match very loosely and probably don't care too much. `v*-jito` matches very loosely already, if we were concerned about restricting patterns we'd probably want something like `v[0-9]+.[0-9]+.[0-9]*-jito` to restrict to pure semantic versions and then any equivalent suffix tightening.

My take is we control tag, so it's probably fine as-is.

PR #1626, opened 2026-09-05
banking_stage: fix BAM tip refresh and simplify scheduler plumbing

esemeniuc:
#### Problem

BAM can admit work with stale tip configuration after builder or commission updates and same-bank reconnects. Repeated configuration cycles within one blockhash can also replay a crank signature, fail with `AlreadyProcessed`, and stall admission. The scheduler carries redundant plumbing and BAM-specific helpers in shared Agave modules.

#### Summary of Changes

- Refresh tip configuration before new admission when the concrete Bank or published metadata snapshot changes, including reconnects that republish equal values. Wait for outstanding admissions to settle, use one snapshot for preparation and caching, and retain queued work if preparation fails. Matching accounts require no crank or additional cost.
- Give required cranks a fresh signature by adding an unused random read-only account, so repeated builder or commission cycles work within the same blockhash.
- Track one BAM batch per work item and validate all stored members before acquiring a reusable work object. Reject incomplete batches without partial execution and unblock dependents when batch state is missing.
- Remove redundant per-work expiry checks, result wrappers, and constructor plumbing while preserving slot-boundary cleanup and ordered cost admission/refunds.
- Avoid rebase conflict: keep cost admission, tip preparation, and BAM response conversion in existing BAM modules.


0xEdgar:
hey @esemeniuc, just landed the master rebase (#1649), please rebase onto master.

0xEdgar (commented):
overall looks fine. i wonder if we can simplify this a ton by apply config changes at the next available slot? this would remove same bank refresh path and snapshot identity tracking. imo the simplicity is worth the latency tradeoff, as long as BAM / BE handoff follows the same rule.

esemeniuc:
> overall looks fine. i wonder if we can simplify this a ton by apply config changes at the next available slot? this would remove same bank refresh path and snapshot identity tracking. imo the simplicity is worth the latency tradeoff, as long as BAM / BE handoff follows the same rule.

applying config changes at the next slot should remove ~18 lines from the bam scheduler according to LLM, not sure if worth redo rn

0xEdgar (approved):

haoranjito (approved):

PR #1649, opened 2026-09-10
Jito Master rebase 09-10-26

0xEdgar:
Rebase onto [Agave master 5edf512403](https://github.com/anza-xyz/agave/commit/5edf5124039477af74ea0c4e039fad35b88e3361):
119 upstream commits, preserving the existing Jito patch.

Pulled in cost-accounting and bank-lifecycle changes, Geyser/SVM API updates,
accounts/rewards/BLS fixes, fastboot startup hints, and RPC/CLI updates.
Build changes include Rust 1.98.1, Tombi, and a separate conformance workspace.

Review carefully:

- **Cost accounting and bank replacement:**
  [qos_service.rs](https://github.com/jito-foundation/jito-solana/blob/f70ec83004ffb04dbbef1521fe5a1786c1a6151f/core/src/banking_stage/qos_service.rs#L62)
  and BAM/bundle callers now pass transaction write sets separately from costs.
  Check rollback/refunds and same-slot replacement bank handling.
- **Geyser plugins:**
  [transaction V4, block V5, and footer V0_0_2](https://github.com/jito-foundation/jito-solana/blob/f70ec83004ffb04dbbef1521fe5a1786c1a6151f/geyser-plugin-interface/src/geyser_plugin_interface.rs#L130)
  replace older payload variants. Downstream plugins need updating.
- **SVM and nonce checks:**
  [account callbacks](https://github.com/jito-foundation/jito-solana/blob/f70ec83004ffb04dbbef1521fe5a1786c1a6151f/svm-callback/src/lib.rs#L37)
  no longer return modification slots. BAM/bundle
  [external prechecks](https://github.com/jito-foundation/jito-solana/blob/f70ec83004ffb04dbbef1521fe5a1786c1a6151f/runtime/src/bank/check_transactions.rs#L69)
  adopt upstream nonce-authority checks; replay settings remain unchanged.
- **Other API changes:** scheduler handshake types move to
  [agave-scheduler-handshake 6.0.0](https://github.com/jito-foundation/jito-solana/blob/f70ec83004ffb04dbbef1521fe5a1786c1a6151f/Cargo.toml#L184),
  receive paths use [BytesPacket](https://github.com/jito-foundation/jito-solana/blob/f70ec83004ffb04dbbef1521fe5a1786c1a6151f/perf/src/packet.rs#L78),
[...]

PR #1650, opened 2026-09-11
Add BAM Discovery Mechanism to BAM Connection Flow

badgerrupt0:
This change introduces `bam_discovery.rs`, which functions as a mechanism for the validator to hit the future upstream BAM Registry endpoint. This registry will provide the validator with a list of healthy BAM nodes in response, from which the validator will probe and sort by RTT time via calling `GetBuilderConfig` for each node. The list of nodes will be cached in the event of a disconnect, and re-selected should a node go down. Re-selection chooses from the lowest RTT for a given validator. The discovery mechanism is designed *only* to swap out the BAM URL from `BamManager` when the connection has failed, or hasn't yet started. 

The goal of this change is to move forward with automating BAM node discovery, and to gradually switch BAM Node <-> Validator connections onto HTTPS, while reducing downtime. `--bam-registry-url` is added as an optional argument that activates this feature. Supplying `--bam-url` is still currently supported, and overrides any `--bam-registry-url` value set.


0xEdgar:
hey @badgerrupt0, just landed the master rebase (#1649), please rebase onto master.

core/src/bam_discovery.rs
```diff
@@ -0,0 +1,759 @@
...
+    /// When the registry built this list.
+    #[serde(default, deserialize_with = "lenient_rfc3339")]
+    generated_at: Option<DateTime<Utc>>,
+    nodes: Vec<ServedNode>,
+}
+
+/// The registry sends RFC 3339. A missing or malformed value costs one metric
+/// rather than failing the whole document.
+fn lenient_rfc3339<'de, D>(deserializer: D) -> Result<Option<DateTime<Utc>>, D::Error>
+where
```
badgerrupt0: Wonder if worth removing this; this was originally added to have some insight on the validator-client side as to how recent the registry's information is, but dropping it would reduce logic and make it more wire-compatible. Thoughts?

core/src/bam_discovery.rs
```diff
@@ -0,0 +1,759 @@
...
+            atomic::{AtomicBool, AtomicU8, Ordering},
+        },
+        thread::{self, Builder, JoinHandle},
+        time::{Duration, Instant},
+    },
+    tokio::time::timeout,
+};
+
+/// How often the published node list is re-read while a session is live.
+const RESYNC_INTERVAL_LIVE: Duration = Duration::from_secs(60);
```
0xEdgar, line 22: thoughts on moving these to a serde struct loadable from file? or do you not envision these ever changing

core/src/bam_discovery.rs
```diff
@@ -0,0 +1,759 @@
...
+#[derive(Clone, Debug, PartialEq, Eq)]
+struct RankedNode {
+    url: String,
+    region: String,
+    rtt_us: u64,
+}
+
+/// The node list a `--bam-url` names, or `None` when it names one node. A node
+/// is a bare host and port; the node list carries a path.
+pub fn registry_url(bam_url: &str) -> Option<&str> {
```
0xEdgar: this function a bit overloaded as written imo; does url parsing, classification ,and silently treating any parse errors as direct node mode . 

suggestion: parse / validate at config boundary, then use `is_registry_url(&Url) -> bool`

core/src/tpu.rs
```diff
@@ -433,6 +435,19 @@ impl Tpu {
         // structurally so changing the shared URL cannot activate BAM in external-scheduler mode.
         let bam_dependencies = scheduler_bindings.is_none().then_some(bam_dependencies);
 
+        // `bam_url` is what the operator asked for, `connect_url` is what
+        // BamManager dials. Discovery resolves one into the other, so it runs
+        // whenever BAM does and inherits the exclusion above.
+        let connect_url = Arc::new(ArcSwap::from_pointee(None));
+        let bam_discovery = bam_dependencies.is_some().then(|| {
```
0xEdgar: do we want discovery to be mandatory even if an operator directly configures the BAM node? or should we init networking lazily when registry is actually needed?

core/src/bam_discovery.rs
```diff
@@ -0,0 +1,759 @@
...
+/// `BamManager` reconnects whenever that url changes.
+pub struct BamDiscovery {
+    /// Background worker that fetches, probes and publishes.
+    thread_hdl: JoinHandle<()>,
+}
+
+impl BamDiscovery {
+    pub fn new(
+        exit: Arc<AtomicBool>,
+        bam_config: Arc<ArcSwap<Option<String>>>,
```
0xEdgar: nit: rename `configured_url` ?

core/src/bam_discovery.rs
```diff
@@ -0,0 +1,759 @@
...
+        let http_client = match reqwest::Client::builder().timeout(FETCH_TIMEOUT).build() {
+            Ok(client) => client,
+            Err(err) => {
+                error!("Failed to build BAM discovery http client, discovery disabled: {err}");
+                return;
+            }
+        };
+
+        // The full published list, kept across fetch failures.
+        let mut nodes: Vec<ServedNode> = Vec::new();
```
0xEdgar: `registry_nodes` ?

core/src/bam_discovery.rs
```diff
@@ -0,0 +1,759 @@
...
+    /// Whether the BAM session is usable. Everything past `Connecting` is
+    /// authenticated, including the unbounded `DrainingBlockEngine` wait.
+    /// `Connecting` is not: BamManager cycles it against `Disconnected` about once
+    /// a second against a node that connects then fails its health check.
+    fn is_live(state: BamConnectionState) -> bool {
+        state as u8 > BamConnectionState::Connecting as u8
+    }
+
+    /// Advance the stall clock by one poll, or clear it once the session is live.
+    fn stall_after_poll(state: BamConnectionState, stalled_for: Duration) -> Duration {
```
0xEdgar: my understanding is this accounting doesn't take into account network latency, meaning the 10 second grace period can take substantially longer when the registry itself is slow to fetch or probe.

core/src/bam_discovery.rs
```diff
@@ -0,0 +1,759 @@
...
+                continue;
+            };
+
+            let state = Self::connection_state(&bam_enabled);
+
+            // Time without a live session, not time in one state.
+            let stuck = stalled_for >= CONNECT_GRACE;
+
+            if time_until_resync == Duration::ZERO {
+                if let Some(served) = runtime.block_on(Self::fetch(&http_client, registry_url))
```
0xEdgar: If the connection to the current BAM node becomes healthy while discovery is waiting on the registry fetch, can we still advance to another candidate because stuck was computed before the wait?

core/src/bam_discovery.rs
```diff
@@ -0,0 +1,759 @@
...
+                // Nothing measured against the old value survives it.
+                nodes.clear();
+                ranked.clear();
+                cursor = 0;
+                time_until_resync = Duration::ZERO;
+                time_until_probe = Duration::ZERO;
+                stalled_for = Duration::ZERO;
+            }
+
+            // Empty stays disconnected, and a single node is already the answer.
```
0xEdgar: what does this comment mean lol

core/src/bam_discovery.rs
```diff
@@ -0,0 +1,759 @@
...
+        let mut stalled_for = Duration::ZERO;
+        let mut config = bam_config.load_full();
+        Self::announce(&config);
+
+        while !exit.load(Ordering::Relaxed) {
+            let current = bam_config.load_full();
+            if current != config {
+                config = current;
+                Self::announce(&config);
+                // Nothing measured against the old value survives it.
```
0xEdgar: what does this comment mean

core/src/bam_discovery.rs
```diff
@@ -0,0 +1,759 @@
...
+    /// Advance the stall clock by one poll, or clear it once the session is live.
+    fn stall_after_poll(state: BamConnectionState, stalled_for: Duration) -> Duration {
+        if Self::is_live(state) {
+            Duration::ZERO
+        } else {
+            stalled_for.saturating_add(POLL_INTERVAL)
+        }
+    }
+
+    /// True when the shared url does not name a node the registry currently
```
0xEdgar: what does this comment mean

core/src/bam_discovery.rs
```diff
@@ -0,0 +1,759 @@
...
+            "bam_discovery-probe_round",
+            ("probed", pool.len() as i64, i64),
+            ("responded", ranked.len() as i64, i64),
+        );
+        ranked
+    }
+
+    /// One row per probed node, responders first in ranking order. Nodes that did
+    /// not answer are absent from the ranking, so they are listed after it.
+    fn probe_table(pool: &[ServedNode], ranked: &[RankedNode]) -> String {
```
0xEdgar: nit(clarity): `format_probe_table` ?

core/src/bam_discovery.rs
```diff
@@ -0,0 +1,759 @@
...
+    pub fn join(self) -> thread::Result<()> {
+        self.thread_hdl.join()
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use {super::*, test_case::test_case};
+
+    /// Byte-for-byte the object the registry pins in its own snapshot test, so a
```
0xEdgar: rm claudese pls

core/src/bam_discovery.rs
```diff
@@ -0,0 +1,759 @@
...
+        let served: ServedNodes = serde_json::from_str(&json).unwrap();
+        assert_eq!(
+            served.generated_at,
+            Some("2026-09-09T12:34:56Z".parse::<DateTime<Utc>>().unwrap()),
+        );
+        assert_eq!(served.nodes.len(), 1);
+    }
+
+    /// The timestamp only feeds a metric, so a broken one must still yield a
+    /// usable node list.
```
0xEdgar:
```suggestion
// An invalid generated_at should not prevent loading the node list.
```

core/src/bam_discovery.rs
badgerrupt0: We could do that if we wanted to allow for this to be tweaked by the operator but I do not see these changing once we find a stable state for them

core/src/tpu.rs
badgerrupt0: Yeah lazy init is the move here, good call out!

core/src/bam_manager.rs
```diff
@@ -165,7 +186,7 @@ impl BamManager {
                     }
 
                     // Try to connect to BAM
-                    let Some(url) = configured_bam_url.as_ref() else {
+                    let Some(url) = connect_url.as_ref() else {
```
esemeniuc, line 189: Once this URL is selected, could the in-flight connection setup listen for BAM configuration changes and stop when the URL is changed or disabled? Connection and health setup can keep using the previous URL for about 16 seconds, so a successful runtime update is not applied promptly and may briefly accept an already stale connection.

can be something like
```
  let outcome = runtime.block_on(async {
      tokio::select! {
          biased;

          _ = wait_for_newer_config(&mut config_rx, attempt_generation) => {
              HealthResult::ConfigChanged
          }

          healthy = connection.wait_until_healthy_and_config_received_async(
              MAX_DURATION_BETWEEN_NODE_HEARTBEATS,
              &exit,
          ) => {
              HealthResult::Finished(healthy)
          }
      }
  });

  if matches!(outcome, HealthResult::ConfigChanged) {
      outbound_receiver = Some(runtime.block_on(connection.shutdown()));
      Self::set_bam_disconnected(&dependencies);
      continue;
  }
```

validator/src/admin_rpc_service.rs
```diff
@@ -647,35 +645,19 @@ impl AdminRpc for AdminRpcImpl {
...
-                (
-                    "previous_bam_url",
-                    old_bam_url.as_ref().clone().unwrap_or_default(),
-                    String
-                ),
-            );
-        }
+        let bam_url = bam_url
+            .map(|url| {
+                crate::commands::bam::normalize_bam_url(&url)
```
esemeniuc, line 657: Could we trim and validate the final URL here?

`setBamUrl("http://localhost:50055 ")` passes `Url::parse`, but the string is rejected by tonic, so the RPC succeeds while BAM cannot connect. previous `Endpoint` check rejected it.

core/src/bam_discovery.rs
```diff
@@ -0,0 +1,837 @@
...
+        // A node can answer the probe and still refuse the scheduler stream.
+        let current_url = bam_url.load_full();
+        if stuck && BamDiscovery::on_pick(current_url.as_deref(), &self.ranked, self.cursor) {
+            self.cursor = BamDiscovery::advance(self.cursor, self.ranked.len());
+            if self.cursor == 0 {
+                self.ranked.clear();
+            }
+        }
+
+        if (stuck || BamDiscovery::needs_pick(current_url.as_deref(), &self.registry_nodes))
```
esemeniuc, line 168: Could grace start when BamManager actually begins trying this URL?

it can still spend ~16s on the previous node after publication: B can be published at t=20, first tried at t=34, and advanced at t=40 before it can become healthy. A connection can also become healthy between the state read above and this publish. Coordinating or cancelling stale setup would avoid both premature switches.

core/src/bam_discovery.rs
```diff
@@ -0,0 +1,837 @@
...
+            .get(cursor)
+            .is_some_and(|node| current_url == Some(node.url.as_str()))
+    }
+
+    fn advance(cursor: usize, len: usize) -> usize {
+        (cursor + 1) % len.max(1)
+    }
+
+    fn publish(bam_url: &ArcSwap<Option<String>>, node: &RankedNode) -> bool {
+        if bam_url.load().as_deref() == Some(node.url.as_str()) {
```
esemeniuc, line 272: This equality load and bool return appear redundant: discovery is the sole writer, `on_pick` advances an equal ranked URL, and `needs_pick` excludes the current URL from the ranking. `publish` can store unconditionally and the caller can reset the timer.

core/src/bam_discovery.rs
```diff
@@ -0,0 +1,837 @@
...
+        });
+
+        Some(Self { selected_url, task })
+    }
+
+    pub fn selected_url(&self) -> Arc<Option<String>> {
+        self.selected_url.load_full()
+    }
+
+    fn registry_url(configured: &Option<String>) -> Option<String> {
```
esemeniuc: This parser is used only by `new`; could we inline the registry check there and drop the helper? Malformed input still reaches the manager connection-error path.

validator/src/commands/bam/mod.rs
```diff
@@ -18,81 +19,90 @@ const DEFAULT_BAM_URL_SCHEME: &str = "http";
...
+    let host_port = authority.rsplit('@').next().unwrap();
+    let port = if host_port.starts_with('[') {
+        host_port
+            .find(']')
+            .and_then(|host_end| host_port[host_end + 1..].strip_prefix(':'))
+    } else {
+        host_port.rsplit_once(':').map(|(_, port)| port)
+    };
+
+    match port {
```
esemeniuc, line 82: The empty-port and omitted-port arms only differ by `:`. Could return early for a nonempty explicit port, then use one formatting path with `:` only when the port is omitted? This preserves `:80`/`:443` and empty-port behavior.

core/src/bam_discovery.rs
```diff
@@ -0,0 +1,837 @@
...
+                    node.url(),
+                    node.region,
+                    "no answer"
+                )
+            });
+
+        std::iter::once(header)
+            .chain(responded)
+            .chain(silent)
+            .collect::<Vec<_>>()
```
esemeniuc, line 416: Could this write rows directly into one `String` with `fmt::Write` instead of collecting row strings and joining? Prefix each appended row with `\n` to preserve the table output; the silent-node loop can also format each URL once.

validator/src/commands/bam/mod.rs
```diff
@@ -12,87 +14,101 @@ pub enum BamUrlError {
...
+            url: url_str.to_owned(),
+            source: ParseError::InvalidPort,
+        });
+    }
+
+    if is_registry_url(&url) {
+        return Ok(parse_target);
+    }
+
+    let authority_start = parse_target.find("://").unwrap() + 3;
```
0xEdgar: suggestion: Move the existing Endpoint::from_shared validation earlier and use its parsed URI from tonic: 

```
endpoint.uri().port_u16() // Preserves explicit 80/443.
endpoint.uri().authority() // Handles hostname and IPv6 parsing.
```

PR #1652, opened 2026-09-12
feat(bam): Add parent identity to LeaderState

0xEdgar:
Populate the `parent_slot`, `parent_block_id` and `atomic_batches_enabled`
fields from jito-labs/bam-protos#18 so BAM can create its bank for the
slot before any shred arrives (MEV-12789, MEV-12835).

- read the working bank and its readiness from one `SharedLeaderState`
  snapshot so a same-slot replacement cannot mix them
- bump `jito-protos/bam-protos` to https://github.com/jito-labs/bam-protos/pull/18

### DIFF
--- a/core/src/bam_manager.rs
+++ b/core/src/bam_manager.rs
@@ -269,11 +269,13 @@ impl BamManager {
                         .store(BamConnectionState::Connected as u8, Ordering::Release);
                 }
             }
-            // Send leader state if we are in a leader slot
-            if let Some(bank) = shared_leader_state.load().working_bank()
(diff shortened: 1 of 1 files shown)


core/src/bam_manager.rs
```diff
@@ -342,6 +344,9 @@ impl BamManager {
             slot: bank.slot(),
             tick: (bank.tick_height() % bank.ticks_per_slot()) as u32,
             slot_cu_budget_remaining,
```
buffalu, line 346: do we need to add the current bank id here too? i need to read into AG more, but think we need to target the bank slot/bank id when were sending the atomic tx batches so that our speculative bank is built properly

PR #1654, opened 2026-09-12
fix(bam): Emit opening header at Alpenglow slot start

0xEdgar:
Fixes MEV-12789. Under Alpenglow, BAM waits for slot N's first shred to
build its bank, and the leader emits no shred until a transaction arrives
or the slot closes. BAM leaders in positions 2–4 never got out of the
deadlock.

When a BAM URL is set, emit the block header right after bank installation instead of letting broadcast prepend it to the first entry batch. Broadcast skips its own header when it sees an explicit one.

Because the parent freezes before broadcast assigns its block ID, the leader waits for both before installing the child, bounded by the block timer. Optimistic starts (fast leader handover) skip the eager header and fall back to the old path. 

Metrics: add  `opening_header_sent` and `parent_block_id_wait_us` to `slot-metrics`

Backport: `v4.3` after `master`, since `v4.3.0-rc.1-jito` is the testnet
activation branch.

**A = existing header timing; B = immediate opening header.** Four
validators, 200 ms slots, fast handover disabled, 50 single-transaction
bundles/sec through BAM, 159 rotations per arm.

Median ms from producer bank installation

| Slot in rotation | A: BNV bank inserted | B: BNV bank inserted | A: first customer forward | B: first customer forward |
| --- | ---: | ---: | ---: | ---: |
| 1 | 125.51 | 2.58 | 126.83 | 3.33 |
| 2 | 201.19 | 7.31 | 202.34 | 8.24 |
| 3 | 201.10 | 7.14 | 202.05 | 7.68 |
| 4 | 201.14 | 7.26 | 202.17 | 7.74 |

Community cluster: TODO once today's run finishes.


core/src/block_creation_loop.rs
```diff
@@ -192,6 +192,8 @@ struct LeaderContext {
     // Metrics
     metrics: LoopMetrics,
 
+    experiment_open_header: bool,
```
0xEdgar: todo: clean up once we decide this is correct approach

core/src/block_creation_loop.rs
```diff
@@ -1263,6 +1278,20 @@ fn start_leader_wait_for_parent_replay_with_used_bytes(
                 replay_behind_measure.stop();
                 slot_metrics.replay_is_behind_us += replay_behind_measure.as_us();
             }
+            Err(StartLeaderError::ParentBlockIdPending(_, _)) => {
+                trace!(
+                    "{my_pubkey}: Attempting to produce slot {slot}, however parent {parent_slot} \
+                     has no block id yet; waiting for broadcast"
+                );
+                // Broadcast assigns the block ID without a notification; poll briefly.
+                let mut wait_measure = Measure::start("parent_block_id_wait");
```
0xEdgar: hmmm

core/src/block_creation_loop.rs
```diff
@@ -159,6 +160,10 @@ pub struct BlockCreationLoopConfig {
 
     /// Sender for packets to banking stage (used to re-inject transactions after sad leader handover).
     pub banking_stage_sender: BankingPacketSender,
+
+    /// BAM URL shared with the admin RPC. Leaders connected to BAM emit an
+    /// opening header at slot start so BAM can build the bank immediately.
+    pub bam_url: Arc<ArcSwap<Option<String>>>,
```
buffalu, line 166: is there any reason to not always do this? idk if we need BAM connected or not flag

core/src/block_creation_loop.rs
```diff
@@ -1422,6 +1471,15 @@ fn create_and_insert_leader_bank(
         .unwrap()
         .set_bank_with_atomic_batches_enabled(tpu_bank, atomic_batches_enabled);
 
+    if let Some(marker) = opening_header {
```
buffalu, line 1474: can you explain why we need this instead of just changing standard broadcast run code?

AshwinSekar (commented):
I'm wondering if we actually need the wait in BCL

Basically as you've written:
1. BCL produces the first leader slot without waiting (as the parent belonging to a prior leader will have the block id set by replay stage)
2. When BCL goes to produce the second leader slot, broadcast might not yet have finished shredding and disseminating the first leader slot, so we wait
3. In both cases at the start we send an explicit header after grabbing the block id

Broadcast will already immediately proceed without coalescing if it receives a block marker
https://github.com/anza-xyz/agave/blob/beee69b9589e0989821ba5750bfc1cd88b4dab9b/turbine/src/broadcast_stage/broadcast_utils.rs#L109-L116

How about instead
1. BCL always sends an explicit header immediately on bank creation
2. This header doesn't have a parent block id set - it's simply a "start the block" event
3. If broadcast sees that the slot has changed and we've received a header, go ahead and fill in the header and send it out

Basically the change would hijack the already existing `maybe_send_header` functionality in broadcast:
https://github.com/anza-xyz/agave/blob/beee69b9589e0989821ba5750bfc1cd88b4dab9b/turbine/src/broadcast_stage/standard_broadcast_run.rs#L418-L425

And in this case ignore the fake header sent by BCL (e.g. remove the construction of `shreds` and this extend)
https://github.com/anza-xyz/agave/blob/beee69b9589e0989821ba5750bfc1cd88b4dab9b/turbine/src/broadcast_stage/standard_broadcast_run.rs#L442


I think this would amount to a less invasive change that's easier to maintain. Also agree with Lucas that we don't need to gate on the bam url - we can just always do this.

esemeniuc (commented):
rm

core/src/block_creation_loop.rs
```diff
@@ -1263,6 +1278,16 @@ fn start_leader_wait_for_parent_replay_with_used_bytes(
                 replay_behind_measure.stop();
                 slot_metrics.replay_is_behind_us += replay_behind_measure.as_us();
             }
+            Err(StartLeaderError::ParentBlockIdPending(_, _)) => {
```
esemeniuc: This wait adds a new failure mode. With 200 ms slots, position 1 reaches its production timeout at approximately 144 ms. The deadline for position 2 is approximately 344 ms. StandardBroadcastRun sets the local parent block ID only after it processes the parent terminal tick and queues its shreds to both downstream outputs.

Before this change, BCL could start a consecutive child while the parent ID was absent. The parent terminal component and the child components use the same ordered path. This path uses bounded queues with 2,048 entries. These queues can hold the small number of components from a sparse slot. When Standard Broadcast resumes, it processes the parent first, sets its block ID, and then processes the child.

If Standard Broadcast does not process the queued parent terminal component before the position-2 deadline, this wait returns ReplayIsBehind. The error at line 680 then exits produce_window, and the validator does not produce positions 2 through 4.

Please keep parent ID resolution in Standard Broadcast. One option is to send a typed internal StartBank event through the existing ordered path. Standard Broadcast can create the real header after it processes the parent terminal component. Avoid using an invalid BlockHeader as the internal control event.

turbine/src/broadcast_stage/standard_broadcast_run.rs
```diff
@@ -887,6 +894,153 @@ mod test {
...
+            &forks,
+        )
+        .unwrap();
+        let header_end = run.next_shred_index;
+        assert_eq!(header_end, DATA_SHREDS_PER_FEC_BLOCK as u32); // Exactly one header FEC set.
+        let meta = blockstore.meta(1).unwrap().unwrap();
+        assert_eq!(meta.parent_slot, Some(parent.slot()));
+        assert_eq!(meta.parent_block_id, parent.block_id().unwrap());
+        assert!(!blockstore.is_full(1));
+        let genesis_marker = GenesisCertBlockMarker::try_from((*genesis_cert).clone()).unwrap();
```
esemeniuc: This test does not construct a block that replay accepts. setup() creates a slot-0 parent. Production does not add a genesis-certificate marker for this parent, and replay rejects this marker with GenesisCertificateInAlpenglowCluster.

Production also sends the footer at max_tick_height - 1 and then sends the final alpentick at max_tick_height. This test marks the footer as final and does not send the required alpentick after the footer. test_process_receive_results only creates, transmits, and inserts shreds. The is_full() result proves that Blockstore received a final shred. It does not prove that replay accepts the component order.

Please omit the genesis marker for this fixture, add the final alpentick after the footer, and replay the completed slot. Alternatively, rename this as a header-deduplication test and remove the production-order claim.

AshwinSekar:
btw I put together what a solution for what my suggestion above would look like. Seems reasonable to make this the default beahvior in Agave as well although we wouldn't backport this to v4.3 https://github.com/anza-xyz/agave/pull/15275

Feel free to copy it

PR #1665, opened 2026-09-17
fast send first shred

haoranjito:
#### Problem

Entry coalescing can delay the first shred of a new slot and replicas' bank creation.

#### Summary of Changes

Pass the broadcaster's current slot into entry reception and return the first entry of a new slot immediately, skipping batch sizing and coalescing setup. Apply this to both broadcast implementations while retaining normal coalescing for subsequent entries and handling slot changes during batching.

#### Testing

Added unit coverage for the first-entry fast path, resumed coalescing, slot changes during batching, and oversized-entry carryover.

Live comparison on `2X5yarNUWGyyiJDtRoPATvJYeNecvWG7zCkBrY2J1to1`: Sep 16 13:30–Sep 17 01:30 UTC before (48 leader rotations), versus Sep 17 02:30–14:30 UTC after (31 rotations).

Median gap from parent-bank freeze to child-bank creation, using the last three slots of each rotation:

| Replica | Before | After | Earlier by |
|---|---:|---:|---:|
| Tokyo 1 | 49.86 ms | 3.22 ms | 46.64 ms |
| Tokyo 2 | 49.54 ms | 3.47 ms | 46.07 ms |
| Frankfurt | 49.53 ms | 3.17 ms | 46.37 ms |
| Newark | 49.45 ms | 4.20 ms | 45.25 ms |

Median gap fell 93.5% (95% rotation-bootstrap interval: 45.6–47.6 ms earlier). Mean reported slot duration was essentially unchanged: 325.35 → 324.12 ms; no throughput improvement was demonstrated.


esemeniuc (dismissed):

turbine/src/broadcast_stage/broadcast_utils.rs
```diff
@@ -143,6 +148,18 @@ fn recv_slot_components_maybe_empty(
         }
     };
 
+    // The first entry of a new slot is sent immediately, so do not size it
```
buffalu, line 151: do we need this if we get the same effect below?

haoranjito: The early return avoids computing serialized_batch_byte_count. 
would save a bit cpu. but it does make the diff larger. 
maybe not worth it?

haoranjito: @esemeniuc

esemeniuc: the early return isn’t needed for correctness, the guard below does the same. 

early return only skips serialization and coalescing setup, the benefit is reduced first entry latency, should save us 0.1us on laptop hardware, maybe 0.15us on server

haoranjito: yeah. so small. maybe not worth optimizing it? keep the change smaller.

esemeniuc (approved):

0xEdgar (approved):

PR #1669, opened 2026-09-17
build: add debug-symbol builds

esemeniuc:
#### Problem

Default `./f` builds lack source-level debug information, and the existing `release-with-debug` mode does not enable Rust frame pointers needed by CPU stack recording.

#### Summary of Changes

- Add `--debug-symbols` as an alias for `--profile release-with-debug`. As with repeated `--profile` options, the last profile selection wins.
- Use one conditional Docker block to preserve embedded DWARF and symbols in both Cargo workspaces and enable Rust/C/C++ frame pointers. Target-specific Rust flags preserve the configured CPU settings, and profile overrides also apply to supported historical checkouts.
- Append `-debug-symbols` to artifact names for this profile and document the option.


dev/Dockerfile
```diff
@@ -106,6 +106,20 @@ esac
 # cargo builds natively (no --target), making this exact by construction.
 TARGET_TRIPLE="$(uname -m)-unknown-linux-gnu"
 
+# Keep debug information embedded when building older refs with --checkout.
+if [ "$BUILD_PROFILE" = "release-with-debug" ]; then
+  export CARGO_PROFILE_RELEASE_WITH_DEBUG_DEBUG=2
+  export CARGO_PROFILE_RELEASE_WITH_DEBUG_STRIP=none
+  export CARGO_PROFILE_RELEASE_WITH_DEBUG_SPLIT_DEBUGINFO=off
```
jito-spaceman, line 113: The three `CARGO_PROFILE_RELEASE_WITH_DEBUG_*` exports are no-ops. Every ref the Dockerfile can build (v3.0.6-jito onward, gated by the `source scripts/agave-build-lists.sh` at line 137) already has `debug=true, strip=false, split-debuginfo="off"`. Older refs abort before cargo runs. Since env beats the manifest, the block would also silently mask any future edit to the profile for `./f` builds only. 

Suggested fix: delete the three lines and the "older refs" comment at 109; drop the matching sentence at dev/BUILD.md:102-103.

dev/Dockerfile
```diff
@@ -106,6 +106,20 @@ esac
 # cargo builds natively (no --target), making this exact by construction.
 TARGET_TRIPLE="$(uname -m)-unknown-linux-gnu"
 
+# Keep debug information embedded when building older refs with --checkout.
+if [ "$BUILD_PROFILE" = "release-with-debug" ]; then
```
jito-spaceman, line 110: Frame pointers only exist in the Docker path, so `release-with-debug` means different binaries from `scripts/cargo-install-all.sh --release-with-debug` or plain `cargo build`. Not sure we use those latter paths anyways, but we should perhaps state that frame pointers are a `./f`-only addition.

jito-spaceman:
Is it worth adding a `--debug-symbols` matrix entry or a dispatch input to CI? Not sure if we need/want to exercise this path.

dev/Dockerfile
```diff
@@ -106,6 +106,20 @@ esac
...
+  export CARGO_PROFILE_RELEASE_WITH_DEBUG_DEBUG=2
+  export CARGO_PROFILE_RELEASE_WITH_DEBUG_STRIP=none
+  export CARGO_PROFILE_RELEASE_WITH_DEBUG_SPLIT_DEBUGINFO=off
+
+  # Target flags merge with .cargo/config.toml, preserving its CPU settings.
+  flags_var="CARGO_TARGET_${TARGET_TRIPLE^^}_RUSTFLAGS"
+  flags_var="${flags_var//-/_}"
+  export "$flags_var=${!flags_var-} -Cforce-frame-pointers=yes"
+  export CFLAGS="${CFLAGS:+$CFLAGS }-g -fno-omit-frame-pointer"
+  export CXXFLAGS="${CXXFLAGS:+$CXXFLAGS }-g -fno-omit-frame-pointer"
```
jito-spaceman, line 120: cc-rs already adds `-g` and `-fno-omit-frame-pointer` when `DEBUG=true`, which this profile sets. All native deps build via cc.

Suggested fix: drop both lines.

dev/Dockerfile
```diff
@@ -106,6 +106,20 @@ esac
...
+# Keep debug information embedded when building older refs with --checkout.
+if [ "$BUILD_PROFILE" = "release-with-debug" ]; then
+  export CARGO_PROFILE_RELEASE_WITH_DEBUG_DEBUG=2
+  export CARGO_PROFILE_RELEASE_WITH_DEBUG_STRIP=none
+  export CARGO_PROFILE_RELEASE_WITH_DEBUG_SPLIT_DEBUGINFO=off
+
+  # Target flags merge with .cargo/config.toml, preserving its CPU settings.
+  flags_var="CARGO_TARGET_${TARGET_TRIPLE^^}_RUSTFLAGS"
+  flags_var="${flags_var//-/_}"
+  export "$flags_var=${!flags_var-} -Cforce-frame-pointers=yes"
```
jito-spaceman, line 118: `${!flags_var-}` and `${CFLAGS:+...}` guard values that cannot exist here?

Suggested fix:
```
triple_env="${TARGET_TRIPLE^^}"
export "CARGO_TARGET_${triple_env//-/_}_RUSTFLAGS=-Cforce-frame-pointers=yes"
```

f
```diff
@@ -294,6 +301,9 @@ fi
 # from --build-dir, not the basename). The target triple is joined with '_'
 # inside the container (e.g. ..._<sha>_x86_64-unknown-linux-gnu).
 artifact_basename="${basename}-${tag//\//_}"
+if [[ "$profile" == release-with-debug ]]; then
+  artifact_basename+="-debug-symbols"
+fi
```
jito-spaceman, line 306: Only one non-release profile gets a suffix. `debug` and `release-with-lto` still collide with release names.

Suggest fix: `[[ $profile != release ]] && artifact_basename+="-${profile}"`, or a small map if `-debug-symbols` is the preferred spelling.

dev/BUILD.md
jito-spaceman, line 1: Suggestion to add with these changes:

`version.yml` currently carries no profile field, and `docker-output/` is replaced on every run; let's add `profile: $BUILD_PROFILE` to the inline manifest here.

dev/Dockerfile
esemeniuc: updated

dev/Dockerfile
esemeniuc: updated

dev/Dockerfile
esemeniuc: updated

dev/Dockerfile
esemeniuc: updated

f
esemeniuc: updated

dev/BUILD.md
esemeniuc: patched the upstream manifest generator to add `profile: release-with-debug`

dev/Dockerfile
```diff
@@ -78,14 +78,6 @@ ARG CI_COMMIT=
 # derived at run time from uname inside the container.
 ARG TARGETARCH
 
-# Promote build-args to env vars so the heredoc-form RUN below sees them.
```
esemeniuc, line 81: removed because stage-scoped `ARG` values already reach subsequent `RUN` instructions and their child processes, including quoted heredocs. Promoting each value to an identical `ENV` entry was redundant, and the final scratch export does not retain builder environment metadata.

dev/Dockerfile
```diff
@@ -175,64 +172,43 @@ fi
...
+  test "$(grep -c '^profile:' scripts/create-release-tarball.sh)" = 0
+fi
 # ----------------------------------------------------------------------------
 
 extra_args=()
 if [ "$INCLUDE_VAL_BINS" = "1" ]; then
   extra_args+=(--include-val-bins)
 fi
 
-if [ "$BUILD_PROFILE" = "release" ]; then
```
esemeniuc, line 199: removed because else block duplicated almost all of scripts/create-release-tarball.sh just to pass a profile flag to cargo-install-all.sh

jito-spaceman (approved):

